PR #51, opened 2016-08-18
Update versioning to require the OMI superproject

jeffaco:
Update versioning to use the [Build-omi](https://github.com/Microsoft/Build-omi) superproject. Due to issues with MI versioning, we can't generate a "bogus" version file if the version file doesn't exist. As a result, this commit will REQUIRE that the super-project be used to build OMI.

This is in preparation for upcoming work to build native installers (both rpm and deb, if possible) as part of the normal OMI non-developer build.

@Microsoft/ostc-devs 
@Microsoft/omi-devs


jeffaco:
Another iteration is coming, you might want to hold off on review.

jeffaco:
Okay, this is fixed to be forgiving if version file does not exist. Please review.

@Microsoft/ostc-devs 
@Microsoft/omi-devs

johnkord:
Signed off

PR #52, opened 2016-08-22
George/faults2

palladia:
Fault parsing enhancements.


Unix/wsman/wsmanparser.c
```diff
@@ -2929,6 +2935,54 @@ int WS_ParseFaultBody(
                     if (XML_Expect(xml, &e, XML_END, PAL_T('s'), PAL_T("FaultDetail")) != 0)
                         RETURN(-1);
                 }
+                else if (Tcscmp(e.data.data, ZT("OMI_Error")) == 0)
+                {
+                    int i;
+                    for (i=0; i<e.attrsSize; ++i)
```
paulcallen: there is an XML_GetAttr or something similar so save having to iterate through it

Unix/xml/xml.c
```diff
@@ -1783,3 +1783,59 @@ int XML_StripWhitespace(
     }
     return 0;
 }
+
+int _ParseCharFault(const XML *self, const XML_Char *data, XML_Char *buffer, size_t buf_size)
```
paulcallen: all xml functions should be prefixed with XML_

Unix/wsman/wsmanclient.c
```diff
@@ -358,6 +367,7 @@ static void _WsmanClient_SendIn_IO_Thread(void *_self, Message* msg)
         if (self->sentResponse)
         {
                 message->result = MI_RESULT_OK;
+                return;
```
paulcallen: leaking message, in fact message is mostly unused as a result

paulcallen: message is leaked in this case

Unix/wsman/wsmanclient.c
```diff
@@ -325,7 +327,6 @@ static MI_Boolean HttpClientCallbackOnResponseFn(
             Message_AddRef(&errorMsg->base);
             Strand_ScheduleAux(&self->strand, PROTOCOLSOCKET_STRANDAUX_POSTMSG);
             PostResultMsg_Release(errorMsg);
-            self->sentResponse = MI_TRUE;
```
paulcallen: this is a result message so you do need to set this

paulcallen:
Fix the above mentioned issues and then it looks good.

PR #56, opened 2016-08-27
Split special case code for shell for request/response

paulcallen:
Added extra code to support receive request

@palladia


palladia:
It looks good.

PR #61, opened 2016-09-02
Disable Travis on MAC OS/X until build issue is resolved

jeffaco:
Travis builds show as failing due to openssl issues on Mac OS/X.

Disable Mac OS/X builds until the situation is resolved.

@Microsoft/ostc-devs 
@Microsoft/omi-devs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ matrix:
     - os: osx
       osx_image: xcode7.3
 
+  allow_failures:
+    - os: osx
+
 addons:
   apt:
     packages:


.travis.yml
```diff
@@ -3,8 +3,8 @@ language: c
 matrix:
   include:
     - os: linux
-    - os: osx
-      osx_image: xcode7.3
+#   - os: osx
```
niroyb: The better way to go is to allow the failures but still build so that we know about them.
https://improveandrepeat.com/2014/01/travis-ci-and-allowed-failures/

niroyb: Signed off with this change

PR #64, opened 2016-09-02
Fix OpenSSL linking on MacOS

### LINKED ISSUE #63: Fix OpenSSL linking on MacOS
OMI (like many other programs) requires the Homebrew version of OpenSSL. We used to follow .NET Core's instructions and `brew link --force` the libraries into the system locations. This was, of course, a terrible idea and https://github.com/homebrew/brew/pull/597 prevents it from being done (for security reasons), which caused our CI build failures for OS X.

andschwa:
Resolves #63.

See commit messages.

@jeffaco I kind of had to guess at the right way to slip this information into OMI's build system (there docs anywhere on how to do this right?). As it is, it works. But I'm afraid `./install` will fail because `buildtool` isn't updated. However, I didn't update `buildtool` because it seems like duplicated logic. `configure` does not use `buildtool` to setup OpenSSL flags, except when it generates `install`; it makes no sense!

Note that if the user of `configure` did not pass any OpenSSL flags at all, and Homebrew's OpenSSL was not found, `configure` exits with 1 as the default "OpenSSL" library (it's not OpenSSL) on OS X does not work with OMI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,6 @@ matrix:
     - os: osx
       osx_image: xcode7.3
 
-  allow_failures:
-    - os: osx
-
 addons:
   apt:
     packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,6 @@ addons:
       - time
 
 before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update && brew install openssl && brew link --force openssl ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update && brew install openssl ; fi
 
 script: cd Unix && ./regress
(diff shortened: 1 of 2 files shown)


jeffaco:
I'd like to discuss this with you. Wouldn't it be nicer to just use the OpenSSL version that Apple ships? They use OpenSSL, so it should be possible to just use theirs rather than install an independent second version.

Let's discuss please, thanks.

andschwa:
@jeffaco Apple does not ship OpenSSL. Apple deprecated OpenSSL in favor of their own library. I completely agree that a better, long-term solution is to rewrite OMI to use Apple's SSL libraries; but that's far more work than we (read: I) can do right now (also, I tried calling you earlier exactly about this 😄 ).

jeffaco:
Signed off. Disappointed that Apple deviated from common usage, will need to look at that later.

andschwa:
@jeffaco oh believe me, we all are :frowning:

PR #66, opened 2016-09-07
Fix builds on Solaris 10 platforms

jeffaco:
Prior OS/X changes for SSL broke the Solaris builds

@Microsoft/ostc-devs 
@andschwa 

PBUILD running successfully, at least for the hosts that are up right now (including Solaris 10 x86, which was previously broken).

### DIFF
--- a/Unix/configure
+++ b/Unix/configure
@@ -1210,7 +1210,7 @@ if [ -z "$openssl" ]; then
 
 fi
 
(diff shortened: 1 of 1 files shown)


andschwa:
Bah I even knew that. I hate shell syntax. Thanks Jeff.

palladia:
Looks good.

PR #69, opened 2016-09-12
New type of fault message

palladia:


Unix/wsman/wsmanclient.c
```diff
@@ -378,12 +383,14 @@ static MI_Boolean ProcessFaultResponse(WsmanClient *self, Page **data)
...
 
+        omiErrorPtr = GetWsmanCimError(errorType, &omiError);
+
         if (self->enumerationState)
         {
             self->enumerationState->endOfSequence = 1;
         }
 
-        PostResult(self, errorMessage, fault.mi_result > 0 ? fault.mi_result : MI_RESULT_FAILED);
+        PostResult(self, errorMessage, fault.mi_result > 0 ? fault.mi_result : MI_RESULT_FAILED, omiErrorPtr);
```
paulcallen: I believe this is a Schedule post on the strand which means there is a chance a different thread could pick it up and deliver to the C API. As a result omiError has gone out of scope from the stack.
Also, the instance was not Constructed (something like InstanceConstruct that creates it in place)

Unix/wsman/wsmanerrorhandling.c
```diff
@@ -1532,10 +1541,22 @@ MI_Result GetWsmanErrorFromSoapFault(
...
         return MI_RESULT_FAILED;
     }
 }
+
+const MI_Instance* 
+GetWsmanCimError(ERROR_TYPES type, OMI_Error *error)
+{
+    if (ERROR_WSMAN_OPERATION_TIMEDOUT == type)
+    {
+        OMI_Error_Set_ProbableCause(error, 111);   // timedout
```
paulcallen: replace magic number with a #define - something that will be more necessary when we have more

Unix/wsman/wsmanclient.c
```diff
@@ -54,15 +55,34 @@ struct _WsmanClient
...
+            errorMsg->errorMessage = Batch_Tcsdup(errorMsg->base.batch, message);
+        }
+        else
+        {
+            errorMsg->errorMessage = NULL;
+        }
         errorMsg->result = result;
+        if (omiError && omiError->ProbableCause.exists)
+        {
+            OMI_ErrorFromErrorCode(errorMsg->base.batch, MI_RESULT_TIME_OUT, MI_T("MI"), MI_T("Timedout"), &newError);
```
paulcallen: We are going to support more than one type of probable cause so I would not hard code this to timeout.
Also, the error type should be MI_RESULT_TYPE_MI.
We can generate the omi_error with strings in them such that we can add the ProbableCauseDecription automatically from the probable cause.
MI_RESULT_TIME_OUT is not a user facing error which means if that is the error code that goes into the mi result value then the user has no clue what it is. It really is an MI_RESULT_FAILED, unfortunately.

Unix/wsman/wsmanclient.c
```diff
@@ -375,12 +402,14 @@ static MI_Boolean ProcessFaultResponse(WsmanClient *self, Page **data)
             errorMessage = fault.mi_message;
         }
 
+        omiErrorPtr = GetWsmanCimError(errorType, &omiError);
```
paulcallen: the use of omiError is not going to work. you are going to have to pass the error information in a different way through to the PostResult -- maybe a dedicated structure that GetWsmanCimError fills in with things like descriptions and errors and probably causes and the likes that the PostResult then just picks up and sends. It is strange to create this error object on the stack and pass it through to the PostResult and it copies one field out into a new object.
We can discuss this more if you want.

PR #80, opened 2016-09-26
Fixes for master pbuild issues. 

yakman2020:


Unix/http/http.c
```diff
@@ -1477,7 +1477,11 @@ static MI_Result _CreateSSLContext(Http* self, const char* sslCipherSuite, Serve
     {
         options |= SSL_OP_NO_SSLv3;
     }
-    if ( SSL_CTX_set_options(sslContext, options) == 0 )
+#if defined(linux)
```
paulcallen: not much point #ifdef-ing this as it hits most platforms. I thought this was just there for solaris or something like that.

yakman2020: Remove the ifdef?

yakman2020: It was there for suse.

yakman2020: ifdef removed. commit amended. 
New comment written, including a note that perhaps we should error out if we don't get the options we requested (ie if the return value anded with options != options).

I didn't just do that because I don't know if that is the expected behavior for the surrounding function.

Unix/http/http.c
```diff
@@ -1477,7 +1477,11 @@ static MI_Result _CreateSSLContext(Http* self, const char* sslCipherSuite, Serve
     {
         options |= SSL_OP_NO_SSLv3;
     }
-    if ( SSL_CTX_set_options(sslContext, options) == 0 )
+
+    // SSL_CTX_set_options is additive (only adding bits in options). Don't bother with call if options == 0.
+    // Todo: we should error out if the return from options doesn't include at least the options passed in, I would expect
```
jeffaco: You'll remove this when you add to feature branch?

yakman2020: In wsman_client_auth (the upcoming auth pr)
    if ( options != 0)
    {

        // If options is zero, the operation is a noop. SSL_CTX_set_options only sets, never clears

        MI_Uint32 set_options = SSL_CTX_set_options(sslContext, options);

        // If we don't get the options we asked for, thats an issue.

        if ((set_options & options) != options);
        {
            trace_SSL_CannotSetOptions( set_options ^ options );
            return MI_RESULT_FAILED;
        }
    }

PR #87, opened 2016-09-30
Fix build problems uncovered by pbuild

yakman2020:
@jeffaco

Deactivates several problematic tests, as indicated in issue section.  adjusts build.mak to remove bash only syntax since other systems use sh.  Modifies configure to deal with the size utility not being available on solaris.

### DIFF
--- a/Unix/build.mak
+++ b/Unix/build.mak
@@ -392,13 +392,20 @@ CHECKDIR=/tmp/omicheck.$(shell ./buildtool username)
 check:
 	rm -rf $(CHECKDIR)
 	( cd $(CONFIGUREDIR); $(OUTPUTDIR)/install --destdir=$(CHECKDIR) )
-	(  export LD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR);  \
-           export DYLD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); \
-           $(CHECKDIR)/$(CONFIG_BINDIR)/omiserver -i -d --livetime 60 --httpport 0 --httpsport 0 --destdir=$(CHECKDIR) )
+	( LD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR);  \
+	  export LD_LIBRARY_PATH;  \
+	  DYLD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR);  \
+	  export DYLD_LIBRARY_PATH; \
+	  $(CHECKDIR)/$(CONFIG_BINDIR)/omiserver -i -d --livetime 60 --httpport 0 --httpsport 0 --destdir=$(CHECKDIR) )
 	sleep 2
-	( export LD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); export DYLD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); $(CHECKDIR)/$(CONFIG_BINDIR)/omicheck --destdir=$(CHECKDIR) )
+	( LD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); export LD_LIBRARY_PATH; \
+	  DYLD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); export DYLD_LIBRARY_PATH; $(CHECKDIR)/$(CONFIG_BINDIR)/omicheck --destdir=$(CHECKDIR) )
 	sleep 2
-	$(CHECKDIR)/$(CONFIG_BINDIR)/omiserver -s --destdir=$(CHECKDIR)
+	( LD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR);  \
+	  export LD_LIBRARY_PATH;  \
+	  DYLD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); \
+	  export DYLD_LIBRARY_PATH; \
+	  $(CHECKDIR)/$(CONFIG_BINDIR)/omiserver -s --destdir=$(CHECKDIR))
 	sleep 2
 	rm -rf $(CHECKDIR)
 
@@ -409,7 +416,7 @@ check:
 ##==============================================================================
 
 size:
-	size $(BINDIR)/omiserver
+	${shell ./buildtool size} $(BINDIR)/omiserver
 
 ##==============================================================================
 ##
--- a/Unix/buildtool
+++ b/Unix/buildtool
@@ -144,6 +144,8 @@ COMMANDS:
         Print command name of C++ compiler.
     ar
         Print command name of archive command.
+    size 
+        Print command name of size command.
     cflags [--debug --pic --errwarn --pal]
         Print C compiler flags.
     cxxflags [--debug --pic --errwarn]
@@ -597,6 +599,52 @@ if [ "$arg1" = "ar" ]; then
             ;;
         *)
             echo ar
+            ;;
+    esac
+
+    exit
+fi
+
+##==============================================================================
+##
+## size command:
+##
+##==============================================================================
+
+if [ "$arg1" = "size" ]; then
+
+    if [ "$argc" != "1" -o "$opts" != "" ]; then
+        echo "Usage: $0 $arg1"
+        echo
+        exit 1
+    fi
+
+    case "$platform" in
+
+        LINUX_IX86_GNU|LINUX_X86_64_GNU)
+            echo size
+            ;;
+        MONTAVISTA_IX86_GNU)
+            echo size
+            ;;
+        NETBSD_IX86_GNU)
+            echo size
+            ;;
+        SUNOS_I86PC_SUNPRO|SUNOS_SPARC_SUNPRO)
+
+            # the size utiltiy is not available on solaris unless you install gnu binutils
+            # separately
+
+            echo "echo \'size not avaliable\'"
+            ;;
+        AIX_PPC_IBM)
+            echo size
+            ;;
+        HPUX_IA64_HP|HPUX_PARISC_HP)
+            echo size
+            ;;
+        DARWIN_IX86_GNU)
+            echo size
             ;;
     esac
 
--- a/Unix/configure
+++ b/Unix/configure
@@ -464,6 +464,7 @@ else
     configuuid="`echo $USER`"
 fi
 
+
 if [ -z "$with_cc" ]; then
     with_cc=`$buildtool cc $buildtoolopts`
 else
@@ -1479,8 +1480,13 @@ cat > $fn <<EOF
 #define CONFIG_HTTPSPORT $httpsport
 #define CONFIG_TMPDIR "$tmpdir"
 #define CONFIG_CIMSCHEMA "$cimschema"
-#define CONFIG_UUID "$configuuid"
 EOF
+
+if [ "$enable_wchar" = "1" ]; then
+    echo "#define CONFIG_UUID L\"$configuuid\"" >>$fn
+else
+    echo "#define CONFIG_UUID \"$configuuid\"" >>$fn
+fi
 
 if [ "$enable_debug" = "1" ]; then
     echo "#define CONFIG_ENABLE_DEBUG" >> $fn
--- a/Unix/tests/codec/mof/blue/parameterset.c
+++ b/Unix/tests/codec/mof/blue/parameterset.c
@@ -21,6 +21,12 @@
 # define STRLEN wcslen
 #else
 #include <wchar.h>
+
+#if !defined(CONFIG_HAVE_WCSCASECMP)
+// defined in pal/strings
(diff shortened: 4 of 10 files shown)


Unix/build.mak
```diff
@@ -392,13 +392,20 @@ CHECKDIR=/tmp/omicheck.$(shell ./buildtool username)
 check:
 	rm -rf $(CHECKDIR)
 	( cd $(CONFIGUREDIR); $(OUTPUTDIR)/install --destdir=$(CHECKDIR) )
-	(  export LD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR);  \
-           export DYLD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); \
+	(  LD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR);  \
+	   export LD_LIBRARY_PATH;  \
+           DYLD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); \
```
jeffaco: Different indentation from prior line? Also, for simplicity, why not just do `DYLD_LIBRARY_PATH=$LD_LIBRARY_PATH`  so a path change need only be done in one place? Also check indentation on line below.

Unix/build.mak
```diff
@@ -392,13 +392,20 @@ CHECKDIR=/tmp/omicheck.$(shell ./buildtool username)
...
+           export DYLD_LIBRARY_PATH; \
            $(CHECKDIR)/$(CONFIG_BINDIR)/omiserver -i -d --livetime 60 --httpport 0 --httpsport 0 --destdir=$(CHECKDIR) )
 	sleep 2
-	( export LD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); export DYLD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); $(CHECKDIR)/$(CONFIG_BINDIR)/omicheck --destdir=$(CHECKDIR) )
+	( LD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); export LD_LIBRARY_PATH; \
+	   DYLD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); export DYLD_LIBRARY_PATH; $(CHECKDIR)/$(CONFIG_BINDIR)/omicheck --destdir=$(CHECKDIR) )
```
jeffaco: Ditto with above.

Unix/build.mak
```diff
@@ -392,13 +392,20 @@ CHECKDIR=/tmp/omicheck.$(shell ./buildtool username)
...
 	sleep 2
-	$(CHECKDIR)/$(CONFIG_BINDIR)/omiserver -s --destdir=$(CHECKDIR)
+	(  LD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR);  \
+	   export LD_LIBRARY_PATH;  \
+           DYLD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); \
+           export DYLD_LIBRARY_PATH; \
+	   $(CHECKDIR)/$(CONFIG_BINDIR)/omiserver -s --destdir=$(CHECKDIR))
```
jeffaco: Ditto. Sometimes I wish Makefiles could have a subroutine of some sort.

Unix/buildtool
```diff
@@ -605,6 +607,48 @@ fi
...
+            echo size
+            ;;
+        MONTAVISTA_IX86_GNU)
+            echo size
+            ;;
+        NETBSD_IX86_GNU)
+            echo size
+            ;;
+        SUNOS_I86PC_SUNPRO|SUNOS_SPARC_SUNPRO)
+            echo echo 'size not avaliable'
```
jeffaco: I think I get what's happening. But a comment to explain what you're doing and perhaps making this line `echo "echo 'size not available'"` might be more clear.

Unix/configure
```diff
@@ -464,6 +464,7 @@ else
     configuuid="`echo $USER`"
 fi
 
+
```
jeffaco: Stray?

Unix/tests/codec/mof/blue/test_lex.cpp
```diff
@@ -36,6 +36,7 @@ using namespace std;
 
 static void TestLexParser(LEX_TEST *test)
 {
+#if !defined(CONFIG_ENABLE_WCHAR) || (defined(linux) || defined(macos))
```
jeffaco: A comment as to why you're doing this `#ifdef` is called for ...

Unix/tests/pal/test_pal.cpp
```diff
@@ -35,7 +35,10 @@
 #include <pal/file.h>
 #include <pal/hashmap.h>
 #include <pal/format.h>
+
+#if (!(defined(linux) || defined(macos)) && !defined(CONFIG_ENABLE_WCHAR))
```
jeffaco: So it looks like this is coupled with similar `#ifdef` statement at line 2422?

You need explanation as to the problem, and perhaps changing `#ifdef` here to be coupled there.  Like perhaps:

```
#if (!(defined(linux) || defined(macos)) && !defined(CONFIG_ENABLE_WCHAR))
#include "test_pal_strings.h"
#else
#define WCHAR_TESTS_DISABLED
#endif
```

Then, at line 2422:

```
#if !defined(WCHAR_TESTS_DISABLED)
```

Then it's more obviouisly coupled.

Unix/buildtool
yakman2020: Done.

Unix/build.mak
yakman2020: These line up perfectly in vim.  Tabs randomizing is sort of hard to avoid in a makefile

yakman2020: Found it. Done.

Unix/build.mak
yakman2020: done

Unix/build.mak
yakman2020: done

Unix/configure
yakman2020: doesn't look that way.
if [ -x "/bin/uuidgen" -o -x "/usr/bin/uuidgen" ]; then

    if [ "$os" != "DARWIN" ]; then
        configuuid="`uuidgen`"
    else
        configuuid="`echo $USER`"
    fi

else
    configuuid="`echo $USER`"
fi

Unix/tests/pal/test_pal.cpp
yakman2020: Good idea. Done.

yakman2020: and a comment to explain the reason.

Unix/build.mak
```diff
@@ -392,13 +392,20 @@ CHECKDIR=/tmp/omicheck.$(shell ./buildtool username)
...
 	rm -rf $(CHECKDIR)
 	( cd $(CONFIGUREDIR); $(OUTPUTDIR)/install --destdir=$(CHECKDIR) )
-	(  export LD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR);  \
-           export DYLD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); \
-           $(CHECKDIR)/$(CONFIG_BINDIR)/omiserver -i -d --livetime 60 --httpport 0 --httpsport 0 --destdir=$(CHECKDIR) )
+	( LD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR);  \
+	  export LD_LIBRARY_PATH;  \
+          DYLD_LIBRARY_PATH=$(LD_LIBRARY_PATH); \
+          export DYLD_LIBRARY_PATH; \
+          $(CHECKDIR)/$(CONFIG_BINDIR)/omiserver -i -d --livetime 60 --httpport 0 --httpsport 0 --destdir=$(CHECKDIR) )
```
jeffaco: This is still not lining up properly, please fix. The problem is that line 396 is tab prefixed and the remainder are space prefixed. If you fix that, things will line up. You may want to use an editor that can display tab stops to make this more clear.

Unix/build.mak
```diff
@@ -392,13 +392,20 @@ CHECKDIR=/tmp/omicheck.$(shell ./buildtool username)
...
-	( export LD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); export DYLD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); $(CHECKDIR)/$(CONFIG_BINDIR)/omicheck --destdir=$(CHECKDIR) )
+	( LD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR); export LD_LIBRARY_PATH; \
+	  DYLD_LIBRARY_PATH=$(LD_LIBRARY_PATH); export DYLD_LIBRARY_PATH; $(CHECKDIR)/$(CONFIG_BINDIR)/omicheck --destdir=$(CHECKDIR) )
 	sleep 2
-	$(CHECKDIR)/$(CONFIG_BINDIR)/omiserver -s --destdir=$(CHECKDIR)
+	( LD_LIBRARY_PATH=$(CHECKDIR)/$(CONFIG_LIBDIR);  \
+	  export LD_LIBRARY_PATH;  \
+          DYLD_LIBRARY_PATH=$(LD_LIBRARY_PATH); \
+          export DYLD_LIBRARY_PATH; \
+	  $(CHECKDIR)/$(CONFIG_BINDIR)/omiserver -s --destdir=$(CHECKDIR))
```
jeffaco: Same here, tab prefixed vs. space prefixed.

Unix/configure
jeffaco: Sorry, by "Stray?", I meant needless newline inserted for no particular reason. A nit, not important.

Unix/tests/codec/mof/blue/parameterset.c
```diff
@@ -21,6 +21,12 @@
 # define STRLEN wcslen
 #else
 #include <wchar.h>
+
+#if !defined(CONFIG_HAVE_WCSCASECMP)
+// defined in pal/strings
+int wcscasecmp(const wchar_t* s1, const wchar_t* s2);
+#endif
```
jeffaco: If this is implemented elsewhere, why do we need the definition here? Can't we grab this from a header file? The manpage for this say to grab definition from `wchar.h`.

Unix/tests/codec/mof/blue/qualifierset.c
```diff
@@ -18,6 +18,12 @@
 # define STRCASECMP _wcsicmp
 #else
 #include <wchar.h>
+
+#if !defined(CONFIG_HAVE_WCSCASECMP)
+// defined in pal/strings
+int wcscasecmp(const wchar_t* s1, const wchar_t* s2);
+#endif
```
jeffaco: See comment from parameterset.c. But I see here that you're including `wchar.h`, and yet you're still defining this. So, obviously, certain platforms don't have this. I see it can depend on POSIX version, which is likely what you're running into.

You can up the posix version on various UNIX platforms, which would likely avoid the need for this. But given that we want to get this committed, it's not worth it at this time. Or, in other words, don't worry about this if you really need it.

Unix/tests/codec/mof/blue/test_lex.cpp
jeffaco: No comment here as to what's going on? Did you just miss it?

Unix/tests/pal/test_pal.cpp
```diff
@@ -2416,6 +2427,7 @@ NitsEndTest
 //
 //==============================================================================
 
+#if WCHAR_TESTS_DISABLED
```
jeffaco: Is this right? Don't you want to DISABLE the test if WCHAR_TESTS_DISABLED, not ENABLE the test? I think you're reversed here.

Unix/tests/pal/test_pal.cpp
```diff
@@ -35,7 +35,18 @@
 #include <pal/file.h>
 #include <pal/hashmap.h>
 #include <pal/format.h>
+
+#if !defined(linux) && !defined(macos) && defined(CONFIG_ENABLE_WCHAR)
```
jeffaco: I'd prefer that you called out where this did NOT work rather than do negative comparisons for where it does work. Something like:

```
#if ( defined(aix) || defined(hpux) || defined(sun) ) && defined(CONFIG_ENABLE_WCHAR)
```

That way, you're specifically calling out where it does NOT work, and it's easier to fix a specific (non-working) platform if it gets fixed with newer compilers.

Unix/build.mak
yakman2020: Done. I hate tabs but you cant get way from them in a makefile

Unix/tests/pal/test_pal.cpp
yakman2020: Fixed

Unix/tests/pal/test_pal.cpp
yakman2020: Done in the same manner as test_pal.cpp

PR #89, opened 2016-10-07
Removes heavy usage of /tmp file system in regress

yakman2020:
Currently build.mak copies then installs the build hierarchy into /tmp file system. That has been changed to omi/tmp.

### DIFF
--- a/Unix/tests/http/test_http.cpp
+++ b/Unix/tests/http/test_http.cpp
@@ -43,7 +43,9 @@ NitsSetup(TestHttpSetup)
 {
     Sock_Start();
     PORT++;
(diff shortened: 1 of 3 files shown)


yakman2020:
@johnkord

Unix/tests/http/test_http.cpp
```diff
@@ -43,7 +43,9 @@ NitsSetup(TestHttpSetup)
 {
     Sock_Start();
     PORT++;
+    IgnoreAuthCalls(1);
```
johnkord: I'm not sure what this does, but the rest is fine with me. I imagine this has to do with PAM stuff. Signed off on everything else!

PR #92, opened 2016-10-13
Paulcallen/send shell

paulcallen:
@yakman2020 @palladia 

Adding 
* Shell client-side support
* Fixed protocol logging to remove dependency on deprecated technology
* Fixed cleanup of wsmanclient (and as a result the httpclient) by fixing the Strand interactions between it and the miapi

### DIFF
--- a/Unix/protocol/protocol.c
+++ b/Unix/protocol/protocol.c
@@ -31,57 +31,6 @@
 #define LOGW2 __LOGW
 #define LOGD2 __LOGD
 #define mistrerror Result_ToString
-typedef struct _NameValuePairs
-{
-    unsigned int Value;
-    const char* Name;
-} NameValuePairs;
-static NameValuePairs MessageTagNames[] =
-{
-    { 1, "GetInstanceReqTag" },
-    { 2, "PostInstanceMsgTag" },
-    { 3, "EnumerateInstancesReqTag" },
-    { 4, "PostResultMsgTag" },
-    { 5, "NoOpReqTag" },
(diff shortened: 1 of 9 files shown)


Unix/protocol/protocol.c
```diff
@@ -21,10 +21,68 @@
...
+    { 25, "ShellDeleteReqTag" },
+    { 26, "ShellReceiveReqTag" },
+    { 27, "ShellSendReqTag" },
+    { 28, "ShellSignalReqTag" },
+    { 29, "ShellConnectReqTag" },
+    { 30, "ShellReconnectReqTag" },
+    { 31, "ShellDisconnectReqTag" },
+    { 32, "ShellCommandReqTag" },
+#endif
+};
```
palladia: There is now a PullRequestTag, #33.

Unix/protocol/protocol.c
```diff
@@ -21,10 +21,68 @@
...
+#define LOGE2 __LOGE
+#define LOGW2 __LOGW
+#define LOGD2 __LOGD
+#define mistrerror Result_ToString
+typedef struct _NameValuePairs
+{
+    unsigned int Value;
+    const char* Name;
+} NameValuePairs;
+static NameValuePairs MessageTagNames[] =
```
palladia: Since MessageTag is an enum, is it really necessary to do NameValuePairs, and not just something like MessageTagNames?

If you really want Name to pair them, it's probably safer to pair MessageTag (as an enum) with the char* strings.

PR #100, opened 2016-11-01
Removed the requirement for linking libgss

yakman2020:
Completely dynalink libgss, no static linking. This relieves the requirement for aix to have the krb5 package installed.

### DIFF
--- a/Unix/http/httpauth.c
+++ b/Unix/http/httpauth.c
@@ -40,7 +40,7 @@ static void _report_error(OM_uint32 major_status, OM_uint32 minor_status, const 
 
 // dlsyms from the dlopen
 
-typedef OM_uint32 KRB5_CALLCONV(*_Gss_Acquire_Cred_With_Password_Func) (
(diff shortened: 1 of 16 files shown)


Unix/http/httpauth.c
```diff
@@ -51,13 +51,107 @@ typedef OM_uint32 KRB5_CALLCONV(*_Gss_Acquire_Cred_With_Password_Func) (
...
+
 typedef enum { NOT_LOADED = 0, LOADING, LOADED } LoadState;
 
 typedef struct _Gss_Extensions {
     LoadState gssLibLoaded;     /* Default is NOT_LOADED */
-    _Gss_Acquire_Cred_With_Password_Func gssAcquireCredwithPassword;
     void *libHandle;
+    // Funcs
 
+    _Gss_Acquire_Cred_With_Password_Func gssAcquireCredwithPassword;
```
jeffaco: Make _ prefix consistent with other func defs.

PR #101, opened 2016-11-01
Add support for Power8

niroyb:
@Microsoft/ostc-devs

### DIFF
--- a/Unix/buildtool
+++ b/Unix/buildtool
@@ -245,12 +245,16 @@ if [ -z "$target" ]; then
             arch=PARISC
             compiler=HP
             ;;
-
         i386:Darwin:*:Darwin* | x86_64:Darwin:*:Darwin*)
             os=DARWIN
             arch=IX86
             compiler=GNU
             ;;
+        ppc*:Linux:*)
+            os=LINUX
+            arch=PPC
+            compiler=GNU
+	    ;;
         *)
             echo "$0: error: unsupported platform: $__m:$__s:$__r:$__v"
             exit 1
@@ -262,7 +266,7 @@ else
 fi
 
 case "$platform" in
-    LINUX_IX86_GNU|LINUX_X86_64_GNU)
+    LINUX_IX86_GNU|LINUX_X86_64_GNU|LINUX_PPC_GNU)
         ;;
     MONTAVISTA_IX86_GNU)
         ;;
@@ -514,7 +518,7 @@ if [ "$arg1" = "cc" ]; then
 
     case "$platform" in
 
-        LINUX_IX86_GNU|LINUX_X86_64_GNU)
+        LINUX_IX86_GNU|LINUX_X86_64_GNU|LINUX_PPC_GNU)
             echo gcc
             ;;
         MONTAVISTA_IX86_GNU)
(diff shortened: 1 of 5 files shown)


Unix/buildtool
```diff
@@ -514,7 +518,7 @@ if [ "$arg1" = "cc" ]; then
 
     case "$platform" in
 
-        LINUX_IX86_GNU|LINUX_X86_64_GNU)
+        LINUX_*)
```
jeffaco: I'm not sure it's a good idea for this to be so "open". Why not just add the platform for Linux on PPC to this list? I think that's safer, FWIW.

Unix/buildtool
```diff
@@ -555,7 +559,7 @@ if [ "$arg1" = "cxx" ]; then
     fi
 
     case "$platform" in
-        LINUX_IX86_GNU|LINUX_X86_64_GNU)
+        LINUX_*)
```
jeffaco: See above.

Unix/buildtool
```diff
@@ -643,7 +647,7 @@ if [ "$arg1" = "size" ]; then
 
             echo "echo \'size not avaliable\'"
             ;;
-        AIX_PPC_IBM)
+        AIX_PPC_IBM|LINUX_PPC_GNU)
```
jeffaco: Ah, that's better. Can't you use these above as well?

Unix/buildtool
```diff
@@ -706,7 +714,7 @@ if [ "$arg1" = "cflags" -o "$arg1" = "cxxflags" ]; then
     r=""
 
     case "$platform" in
-        LINUX_IX86_GNU|LINUX_X86_64_GNU|MONTAVISTA_IX86_GNU|NETBSD_IX86_GNU)
+        LINUX_*|MONTAVISTA_IX86_GNU|NETBSD_IX86_GNU)
```
jeffaco: So ditto for lines 717, 891, 1015, 1099, 1143, and a bunch of others. I won't mention the others, you can find them.

Unix/buildtool
```diff
@@ -725,8 +733,13 @@ if [ "$arg1" = "cflags" -o "$arg1" = "cxxflags" ]; then
             r="$r -fno-strict-aliasing"
             r="$r -D_GNU_SOURCE"
             r="$r -D_XOPEN_SOURCE=600"
-            r="$r -D_BSD_SOURCE"
 
+	    # _BSD_SOURCE is deprecated on PPC
+            if test "$platform" = "LINUX_PPC_GNU"; then
+		r="$r -DEFAULT_SPOURCE"
```
jeffaco: DEFAULT_SPORCE? Should this be DEFAULT_SOURCE?

Unix/configure
```diff
@@ -1789,7 +1789,7 @@ fi
 
 cat $root/common/MI.h >> $fn
 
-echo "Created $fn"
+echo "created $fn"
```
jeffaco: Why this change?

Unix/sock/sock.c
```diff
@@ -361,6 +361,7 @@ MI_Result Sock_Connect(
 
     r = connect(self, &addr->u.sock_addr, (size_t)addr->sock_addr_size);
     connectErrno = errno;
+    (void) connectErrno;
```
jeffaco: connectErrno isn't being used? Then why not get rid of line 363?

Unix/configure
niroyb: All the others are lowercase in this file.

Unix/sock/sock.c
niroyb: It is used if debug logging is compiled in. Line 375

jeffaco: Doesn't this need to be #ifdef'ed for DEBUG or something? I thought that's what we discussed yesterday ...

PR #102, opened 2016-11-02
Fix gss library data symbols

yakman2020:
This fixes an issue causing the client to always fail in negotiated authentication. 
There is also detabbing in a few files.

### DIFF
--- a/Unix/http/http.c
+++ b/Unix/http/http.c
@@ -278,11 +278,11 @@ static MI_Result _Sock_ReadAux(
             handler->acceptDone = MI_TRUE;
             return _Sock_ReadAux(handler,buf,buf_size,sizeRead);
         }
-	else
+    else
         {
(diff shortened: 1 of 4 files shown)


yakman2020:
@Microsoft/ostc-devs

Unix/http/http.c
```diff
@@ -278,11 +278,11 @@ static MI_Result _Sock_ReadAux(
...
         }
-	else
+    else
         {
-	    SSL_load_error_strings(); // registers the error strings for all libcrypto and libssl functions.
-	    SSL_get_error(handler->ssl, res);
-	}
+        SSL_load_error_strings(); // registers the error strings for all libcrypto and libssl functions.
+        SSL_get_error(handler->ssl, res);
+    }
```
jeffaco: If tabs are removed from this file, why is the indentation weird? See 282 and 283, above.

PR #106, opened 2016-11-04
Secure means secure transport

palladia:
Confusion on what the 3rd argument of HttpClient_New_Connect2() mean.  Created new secureTransport variable.

Added new tests to verify "--encryption" parameter of omicli.


palladia:
Review @omi-devs

PR #107, opened 2016-11-08
SSL init only once, stabilize failed connections

paulcallen:
ssl initialization only needs to be called once per process. doing it
more than once can mess itself up or other threads that are using ssl.
fixed couple of issues when connecting to a bad destination... still not
so great, but getting better.

### DIFF
--- a/Unix/miapi/InteractionProtocolHandler.c
+++ b/Unix/miapi/InteractionProtocolHandler.c
@@ -521,7 +521,10 @@ static void InteractionProtocolHandler_Operation_Strand_Finish( _In_ Strand* sel
 //TODO: At some point we need to make sure the thread is shutdown properly without causing deadlocks
 //    Thread_Join(&operation->protocolRunThread, &returnCode);
 
-    Message_Release(&operation->req->base);
+    if (operation->req)
+    {
+        Message_Release(&operation->req->base);
+    }
     PAL_Free(operation);
 }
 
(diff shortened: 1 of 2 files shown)


paulcallen:
@palladia @yakman2020 

Unix/miapi/InteractionProtocolHandler.c
```diff
@@ -521,7 +521,8 @@ static void InteractionProtocolHandler_Operation_Strand_Finish( _In_ Strand* sel
 //TODO: At some point we need to make sure the thread is shutdown properly without causing deadlocks
 //    Thread_Join(&operation->protocolRunThread, &returnCode);
 
-    Message_Release(&operation->req->base);
+    if (operation->req)
+        Message_Release(&operation->req->base);
```
jeffaco: Please use brackets to make clear use of the "if". I think you want:

```
    if (operation->req)
    {
        Message_Release(&operation->req->base);
    }
```

yakman2020: +1.
The program "indent" can do that automatically.

yakman2020: Do you want to ref count this, or just assume it is closed properly at exit?

paulcallen:
updated based on feedback
@palladia @yakman2020 @jeffaco 

PR #117, opened 2016-11-14
George/invalid cred

palladia:


Unix/http/httpclient.c
```diff
@@ -713,10 +713,15 @@ static Http_CallbackResult _ReadHeader(
         rslt = HttpClient_IsAuthorized(handler);
     }
 
-    if (handler->isAuthorized && PRT_CONTINUE != rslt) 
+    HttpClient* self = (HttpClient*)handler->base.data;
+        
+    if (!handler->isAuthorized && !handler->authorizing)
+    {
+        (*self->callbackOnStatus)(self, self->callbackData, MI_RESULT_ACCESS_DENIED, NULL);
```
paulcallen: don't you need to return FALSE as well?

PR #118, opened 2016-11-14
Add --enable-microsoft to configure, enable native installation kits

jeffaco:

First stab at OMI building it's own packages. This is tested on Linux and appears to work fine, running a pbuild now for platform breadth.

Note that this shelveset is designed to work in concert with SCXcore changes as well, review that as well please.

NOTE: After this, I'll make other (unrelated) changes, like `./configure --dev` actually configuring the way that our standard build will, so the appropriate tests are run by regress. But that's separate to OMI building it's own packages, so I wanted to make that a separate commit.

### DIFF
--- a/Unix/configure
+++ b/Unix/configure
@@ -259,6 +259,32 @@ do
         exit 0
         ;;
 
+    --enable-microsoft)
+      enable_preexec=1
+      prefix=/opt/omi
+      localstatedir=/var/opt/omi
+      sysconfdir=/etc/opt/omi/conf
+      certsdir=/etc/opt/omi/ssl
+      enable_native_kits=1
+      ;;
+
+    --enable-native-kits)
+      enable_native_kits=1
+      ;;
+
+    --enable-ulinux)
+      if [ "`uname -s`" != "Linux" ]; then
+          echo "The --enable-ulinux option is only valid on the Linux platform."
+          exit 1
+      fi
+
+      enable_ulinux=1
+      ;;
+
+    --disable-makefile-gen)
+      disable_makefile_gen=1
+      ;;
+
     *)
       echo "$0: unknown option:  $opt"
       exit 1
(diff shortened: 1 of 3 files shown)


Unix/configure
```diff
@@ -361,6 +395,89 @@ buildtool="$root/buildtool"
...
+    esac
+
+    openssl098_cflags=`PKG_CONFIG_PATH=$openssl098_libdir/pkgconfig $pkgconfig --cflags openssl`
+    openssl098_libs=`PKG_CONFIG_PATH=$openssl098_libdir/pkgconfig $pkgconfig --libs openssl`
+    openssl100_cflags=`PKG_CONFIG_PATH=$openssl100_libdir/pkgconfig $pkgconfig --cflags openssl`
+    openssl100_libs=`PKG_CONFIG_PATH=$openssl100_libdir/pkgconfig $pkgconfig --libs openssl`
+
+    echo "checking for local installation of openssl version 0.9.8 on system ..."
+    if [ ! -d "$openssl098_dir" ]; then
+	echo "did not find ${openssl098_dir}, which is required."
```
paulcallen: use spaces instead of tabs

Unix/configure
```diff
@@ -2231,20 +2330,41 @@ echo "created $fn"
...
-export DATADIR=$datadir
 export ROOT=$root
+export ENABLE_ULINUX=$enable_ulinux
+
+OMI_CONFIGURE_QUALS=--enable-microsoft --disable-makefile-gen
 
 all:
+ifeq (\$(ENABLE_ULINUX),1)
+
+	@echo '========================= Performing Building OMI (0.9.8)'
```
paulcallen: can you say ssl version 0.9.8

Unix/configure
jeffaco: Changed to spaces rather than tabs.

Unix/configure
jeffaco: Okay, changed to `Performing Building OMI (SSL 0.9.8)`. Similar change for SSL 1.0.0.

PR #122, opened 2016-11-18
2 changes - omiserver shutdown in tests, compression soap header

paulcallen:
terminate actual omiserver pid in tests 
add support for compression soap header in client 


Unix/tests/cli/test_cli.cpp
```diff
@@ -339,10 +343,45 @@ static int StopServerSudo()
...
+        int status;
+        int numWaits = 0;
+
+        if (PIDFile_Read(&pid) != 0)
             return -1;
+
+        pidStr << pid;
+
+        argv[args++] = sudoPath;
+        argv[args++] = "-E";
```
jeffaco: The -E qualifier doesn't exist on non-Linux platforms, so this won't work.

What we do in SCX is something like this:

Generate a temporary file with shell-sourceable lines to set env for specific items

Then we execute a command script with something like:

. env-script
command-to-execute

The env script will contain something like:

MYVAR1=value-of-MYVAR1
MYVAR2=value-of-MYVAR2

You can write this sort of thing from shell with something like:

echo "MYVAR1=$MYVAR1" >> env-script

Let me know if you have questions on this. But as is, this will definitely fail on non-Linux platforms. A PBUILD should show this.

Unix/tests/cli/test_cli.cpp
```diff
@@ -339,10 +343,45 @@ static int StopServerSudo()
...
+        argv[args++] = "kill";
+        argv[args++] = "-term";
+        argv[args++] = pidStr.str().c_str();
+        argv[args++] = NULL;
+
+        if (Process_StartChild(&killProcess, sudoPath, (char**)argv) != 0)
+            return -1;
+        
+try_again:
```
jeffaco: I'm not wild on `goto` when a `while` loop would have done it. But I know that this is done commonly in OMI ...

Unix/create_user.sh
```diff
@@ -0,0 +1,52 @@
...
+if [ "x$username" = "x" ]; then
+    echo "Creating new user..."
+    username=omi_$USER
+fi
+
+egrep "^$username" /etc/passwd >/dev/null
+if [ $? -eq 0 ]; then
+    echo "User $username exists!"
+else
+    useradd=`which useradd`
```
jeffaco: Isn't the location of useradd consistent across platforms (probably /usr/sbin)? Is there any system where this is NOT the case?

Also, what about Mac? Don't you want to bypass this on Mac OS?

Unix/create_user.sh
```diff
@@ -0,0 +1,52 @@
...
+    if [ "x$useradd" = "x" ]; then
+        echo "Failed to find 'useradd'"
+        exit 1
+    fi
+
+    sudo $useradd -m $username
+    [ $? -eq 0 ] && echo "User $username has been added to system!" || echo "Failed to add a user!"
+fi
+
+userpasswd=`cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 28 | head -n 1`
```
jeffaco: urandom? Isn't random enough?

Unix/create_user.sh
```diff
@@ -0,0 +1,52 @@
...
+if [ "x$userpasswd" = "x" ]; then
+    echo "Failed to generate password"
+    exit 1
+fi
+
+echo $username:$userpasswd | sudo chpasswd
```
jeffaco: chpasswd is Linux-specific, I believe. What about other platforms?

Unix/create_user.sh
```diff
@@ -0,0 +1,52 @@
...
+[ $? -eq 0 ] && echo "User password has been changed!" || echo "Failed to change password!"
```
jeffaco: If failures occur, you just continue in-line. Is that really what you want to do here? Shouldn't we abort if, for example, the account could not be created?

Unix/delete_user.sh
```diff
@@ -0,0 +1,25 @@
...
+fi
+
+egrep "^$username" /etc/passwd >/dev/null
+if [ $? -eq 0 ]; then
+    userdel=`which userdel`
+    if [ "x$userdel" = "x" ]; then
+        echo "Failed to find 'userdel'"
+        exit 1
+    fi
+    sudo $userdel --force $username
```
jeffaco: Do you need a qualifier to delete the home directory? And does a group get created for the user? If so, is that cleaned up?

Unix/delete_user.sh
```diff
@@ -0,0 +1,25 @@
...
+    echo "Deleted user $username"
+
+fi
+
+ntlm_file=$HOME/.ntlm
+
+if [ -f $ntlm_file ]; then
```
jeffaco: You don't need this `if` test. The `rm -f` command is silent if the file doesn't exist:

```
jeffcof:omi> rm -f foobar
jeffcof:omi> echo $?     
0
jeffcof:omi>
```

palladia:
omi-devs please review

Unix/tests/cli/test_cli.cpp
```diff
@@ -274,6 +305,45 @@ static int StartServerSudo()
     argv[args++] = "--loglevel";
     argv[args++] = Log_GetLevelString(Log_GetLevel());
     argv[args++] = NULL;
+#else
+    string executeFile = home;
+    executeFile += "/omi_execute.sh";
```
paulcallen: shouldn't this be in the output/tmp directory, not home.

Unix/tools/auth_tests_cleanup.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+ntlm_file=$HOME/.ntlm
```
paulcallen: if this is coming and going with tests then it should go in the output/tmp directory

Unix/tools/auth_tests_setup.sh
```diff
@@ -0,0 +1,40 @@
...
+#
+# Used by regress and 'make tests'.  Should be run at omi/Unix level.
+
+username=$OMI_USER
+userpasswd=$OMI_PASSWORD
+
+#
+# Clean up previous (if any) running omiserver
+#
+pidFile=./output/var/run/omiserver.pid
```
paulcallen: this is going to cause problems as under regress it uses different directory names per configuration it runs

Unix/tools/auth_tests_setup.sh
```diff
@@ -0,0 +1,40 @@
...
+#
+# if required env variables don't exist, then quit
+#
+if [ "x${username}" != "x" -a "x${userpasswd}" != "x" ]; then
+    SUDO_PATH=`which sudo`
+    export SUDO_PATH
+
+## NTLM setup
+
+    ntlm_file=$HOME/.ntlm
```
paulcallen: tmp file should go in output directory

Unix/tests/cli/test_cli.cpp
```diff
@@ -337,15 +407,57 @@ static int StopServerSudo()
...
+        {
+            if (errno == ECHILD)
+                goto cleanup;
+
+            // EINTR is common in waitpid() and doesn't necessarily indicate failure.
+            //       Give the system 10 times to see if the pid cleanly exits.
+            if (10 < numWaits && errno == EINTR)
+            {
+                numWaits++;
+                Sleep_Milliseconds(10);
```
jeffaco: Is 10 milliseconds (really 10*10 milliseconds, or 100 milliseconds) enough, even on our really slow HP systems? Perhaps, if UNIX, make it longer, like 25 tries (1/2 second)?

If the process finishes quickly, this change will have no effect. It'll only matter if the system is slow and takes longer than you anticipated ...

Unix/build.mak
```diff
@@ -161,10 +161,15 @@ distclean:
 ##==============================================================================
 
 ifdef BUILD_TESTS
+
+.PHONY: tests
+
 tests:
-	( LD_LIBRARY_PATH=$(LIBDIR); export LD_LIBRARY_PATH; DYLD_FALLBACK_LIBRARY_PATH=$(LIBDIR); export DYLD_FALLBACK_LIBRARY_PATH; DYLD_LIBRARY_PATH=$(LIBDIR); export DYLD_LIBRARY_PATH; $(BINDIR)/nits -file:$(TMPDIR)/nitsargs.txt )
+	bash -c "trap 'trap - SIGINT SIGTERM ERR; ./tools/auth_tests_cleanup.sh' SIGINT SIGTERM ERR; ./tools/make_tests.sh $(LIBDIR) $(BINDIR) $(TMPDIR)"
```
jeffaco: I observed (in a later change) that `kill -SIGTERM` doesn't work on Solaris. Does `trap - SIGTERM` work there reliably?

Unix/tools/auth_tests_setup.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/sh
+#
+# Used by regress and 'make tests'.  Should be run at omi/Unix level.
```
jeffaco: Worth doing a basename on the current dir and validating that we are truly at the `omi\Unix` level?

Unix/tools/auth_tests_setup.sh
```diff
@@ -0,0 +1,43 @@
...
+# Clean up previous (if any) running omiserver
+#
+for outputdir in output*
+do
+    pidFile=$outputdir/var/run/omiserver.pid
+    if [ -f $pidFile ]; then
+        pid=`cat $pidFile`
+
+        if [ "x${pid}" != "x" ]; then
+            sudo kill -15 $pid
```
jeffaco: Seems risky. How do you know that some other process hasn't reused the pid, which can happen? This is particularly important since you're using sudo to do the kill, which allows you to kill anything on the system.

Normally, we'll do something like run `ps` looking for the pid and verifying that the process name is what you expect. That's pretty standard for this sort of thing.

Unix/tests/cli/test_cli.cpp
palladia: Will change to 25.

Unix/build.mak
palladia: Sun does recognize SIGTERM, it's just that kill command doesn't support the kill -s SIGTERM syntax.  kill -15 works.

Unix/tools/auth_tests_setup.sh
palladia: Will change to check that process name contains "omiserver".

palladia:
@jeffaco , @paulcallen : Please re-review.

Unix/tests/cli/test_cli.cpp
```diff
@@ -207,32 +214,61 @@ static int StartServer()
...
+    const char* argv[MAX_SERVER_ARGS];
+
+    std::string pamScript = OMI_GetPath(ID_PREFIX);
+    pamScript += "/output/installpam";
+
+    argv[args++] = sudoPath;
+    argv[args++] = pamScript.c_str();
+    argv[args++] = NULL;
+
+    int r = Process_StartChild(&serverProcess, sudoPath, (char**)argv);
```
paulcallen: don't you want to wait for it to complete?

paulcallen:
also, what is causing this test to fail on travis on mac?
	          libtest_http.dylib:TestHttpClient_ChunkedResponse...
test_httpclient.cpp(990): 
	Assert: (s_contentReceivedFromServer) == (string("Test"))
		()
	[Failed]  libtest_http.dylib:TestHttpClient_ChunkedResponse	0.005927

Unix/tests/cli/test_cli.cpp
palladia: Added wait.

PR #129, opened 2016-11-22
Fixes to httpclient for authorization bypass

johnkord:
@Microsoft/ostc-devs

### DIFF
--- a/Unix/http/httpclient.c
+++ b/Unix/http/httpclient.c
@@ -2552,7 +2552,7 @@ MI_Result HttpClient_StartRequest(
            }
            else 
            {
-	       tmp_headers[j] = (char *)headers->data[i];
+               tmp_headers[j] = (char *)headers->data[i];
                j++;
            }
        }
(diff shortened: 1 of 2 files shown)


Unix/http/httpclient.c
```diff
@@ -2551,9 +2552,13 @@ MI_Result HttpClient_StartRequest(
            }
            else 
            {
+	       tmp_headers[j] = (char *)headers->data[i];
+               j++;
```
jeffaco: Indentation difference

Unix/http/httpclient.c
```diff
@@ -2526,19 +2526,20 @@ MI_Result HttpClient_StartRequest(
...
-   const size_t      AUTH_HDR_LEN         = MI_COUNT(AUTH_HDR);
+   const size_t      AUTH_HDR_LEN         = MI_COUNT(AUTH_HDR) - 1;
 
-   if (headers) {
-
-       extra_headers.size = headers->size;
-       extra_headers.data = (const char **)PAL_Malloc(sizeof(char *)*extra_headers.size);
+   if (headers)
+   {
+       tmp_headers = (headers->size != 0) ? (const char **)PAL_Malloc(sizeof(char *) * headers->size) : NULL;
```
jeffaco: I don't think this is getting free'd anywhere. Memory leak, I think.

PR #130, opened 2016-11-23
Add NTLM auth setup script for Ubuntu16

niroyb:
@Microsoft/ostc-devs

### DIFF
--- a/Unix/scripts/installntlm-ubuntu16
+++ b/Unix/scripts/installntlm-ubuntu16
@@ -1,0 +1,177 @@
+#!/bin/bash
+
+# This scripts installs the dependencies and configures SAMBA to enable NTLM authentication
+# and locally cache the credentials.
+# It takes no parameters but requires user interaction for the password prompts.
+# It also requires the omi debian package to be present in the home directory. (See below)
+
+OMI_DEB=$(readlink -f ~/omi-1.1.0-2.rhel.x64.deb)
+POWERSHELL_DEB=$(readlink -f ~/powershell_6.0.0-alpha.12-1ubuntu1.16.04.1_amd64.deb)
+PSRP_DEB=$(readlink -f ~/psrp-1.0.0-0.universal.x64.deb)
+
+# This script automates setting up an ubuntu 16 machine for testing ntlm
+if [ "$EUID" = "0" ]; then
+   echo "This script should NOT be run as root" 1>&2
+   exit 1
+fi
+
+#set -e
+#set -x
+
+is_installed()
+{
+    if [ -z "$1" ]; then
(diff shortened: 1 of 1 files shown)


Unix/scripts/installntlm-ubuntu16
```diff
@@ -0,0 +1,172 @@
+#!/bin/bash
+
+OMI_DEB=$(readlink -f ~/omi-1.1.0-2.rhel.x64.deb)
```
jeffaco: readlink doesn't work on UNIX platforms; I guess since this is specific to Ubuntu 16 (based on it's name), that doesn't matter?

We have other scripts that do this sort of thing in a more transparent way, just FYI ...

Unix/scripts/installntlm-ubuntu16
```diff
@@ -0,0 +1,172 @@
+#!/bin/bash
```
jeffaco: I'd like some comments at the very beginning on what, exactly, this script is for. I'm not asking for a book, just a few lines.

Does the script take any parameters? If so, you should have usage. But if not, I wouldn't mine actually echoing what the script does and asking if we should proceed? This scripts makes a lot of setup changes on your machine, and there's no easy way to "undo" it. As such, if this was run inadvertently, it would be nice to have a prompt of what the script does (again, just a few lines), and then ask "Proceed [No] ?", default No, take Y or YES or y or yes or whatever as affirmative.

While the prompt would be nice, minimally AT LEAST include comments as to what is going on.

Will sign off, do what you feel is right, I prefer the prompt though. Also, was this checked in as executable?

Unix/scripts/installntlm-ubuntu16
```diff
@@ -0,0 +1,172 @@
...
+    sudo net groupmap add ntgroup="Domain Users" unixgroup=users rid=513
+    sudo useradd -m -g users -G administrator administrator
+    sudo pdbedit -a -u administrator
+    sudo service winbind restart
+    wbinfo -i administrator
+    sudo smbpasswd -a administrator
+    wbinfo -K administrator
+fi
+
+echo "# Done"
```
jeffaco: Why echoing # Done (instead of "Done"). Is the # significant in some way?

Unix/scripts/installntlm-ubuntu16
niroyb: There is a lot of Ubuntu specific code here. Maybe the next step will be to get it cross plat.

Unix/scripts/installntlm-ubuntu16
niroyb: It separates it from the output of the other commands.

PR #134, opened 2016-11-30
create version of libmi that has rpath=$ORIGIN

paulcallen:
... for mac and linux, then for installbuilder, copy that version to production server if it exists
@Microsoft/omi-devs

### DIFF
--- a/Unix/installbuilder/GNUmakefile
+++ b/Unix/installbuilder/GNUmakefile
@@ -142,9 +142,15 @@ else # ifneq ($(PF),Darwin)
 
 endif # ifneq ($(PF),Darwin)
 
-	# We also need a raw copy of libmi shared library
+	# We also need a raw copy of libmi shared library. If libmi_origin exists
+	# use that and rename it libmi.
 	mkdir -p $(PACKAGE_DIR)
+ifeq ($(shell test -e $(OUTPUTDIR)/lib/libmi_origin.* && echo -n yes),yes)
+	cp $(OUTPUTDIR)/lib/libmi_origin.* $(PACKAGE_DIR)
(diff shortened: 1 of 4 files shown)


Unix/installbuilder/GNUmakefile
```diff
@@ -142,9 +142,15 @@ else # ifneq ($(PF),Darwin)
 
 endif # ifneq ($(PF),Darwin)
 
-	# We also need a raw copy of libmi shared library
+	# We also need a raw copy of libmi shared library. If libmi_origin exists
+	# use that and rename it libmi.
 	mkdir -p $(PACKAGE_DIR)
-	cp $(OUTPUTDIR)/lib/libmi.* $(PACKAGE_DIR)
+	ifqu ($(shell test -e $(OUTPUTDIR)/lib/libmi_origin.* && echo -n yes),yes)
```
jeffaco: ifqu?  Do you mean `ifeq`?

Unix/midll/origin/GNUmakefile
```diff
@@ -0,0 +1,22 @@
...
+
+DEFINES = MI_CONST= DETECT_FORBIDDEN_FUNCTIONS
+LIBRARIES = miapi protocol sock wql base $(PALLIBS) omi_error wsman http xmlserializer xml micodec mofparser
+
+EXPORTS=../libmi.exp
+
+include $(TOP)/mak/rules.mak
+
+# Setting this after rules as we are overriding the default one we use
+LIBPATHFLAGS = $(shell $(BUILDTOOL) libpath \$$ORIGIN )
```
jeffaco: Are there other libpath flags that may be getting unset inadvertently? Or is this only used for this single purpose?

paulcallen: it seems that it is only having the rpath setting

Unix/installbuilder/GNUmakefile
paulcallen: oh

PR #136, opened 2016-12-02
Fix rpath issues on mac for libmi.dylib

paulcallen:
@Microsoft/omi-devs

### DIFF
--- a/Unix/installbuilder/GNUmakefile
+++ b/Unix/installbuilder/GNUmakefile
@@ -133,25 +133,26 @@ ifneq ($(PF),Darwin)
 
 	mkdir -p $(PACKAGE_DIR)
 	cp $(OUTPUTDIR)/release/omi-* $(PACKAGE_DIR)
+endif # ifneq ($(PF),Darwin)
 
-else # ifneq ($(PF),Darwin)
-
+ifeq ($(PF),Darwin)
 	# We need a raw copy of omicli (for testing on the Mac platform)
 	mkdir -p $(PACKAGE_DIR)
 	cp $(OUTPUTDIR)/bin/omicli $(PACKAGE_DIR)
 
-endif # ifneq ($(PF),Darwin)
-
-	# We also need a raw copy of libmi shared library. If libmi_origin exists
-	# use that and rename it libmi.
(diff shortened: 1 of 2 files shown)


Unix/installbuilder/GNUmakefile
```diff
@@ -133,25 +133,26 @@ ifneq ($(PF),Darwin)
 
 	mkdir -p $(PACKAGE_DIR)
 	cp $(OUTPUTDIR)/release/omi-* $(PACKAGE_DIR)
+endif # ifneq ($(PF),Darwin)
 
-else # ifneq ($(PF),Darwin)
-
+ifeq ($(PF),Darwin)
```
jeffaco: I'm not sure what the difference is here, or was this just a cosmetic change for you?

Unix/installbuilder/GNUmakefile
```diff
@@ -133,25 +133,26 @@ ifneq ($(PF),Darwin)
...
 	mkdir -p $(PACKAGE_DIR)
 	cp $(OUTPUTDIR)/bin/omicli $(PACKAGE_DIR)
 
-endif # ifneq ($(PF),Darwin)
-
-	# We also need a raw copy of libmi shared library. If libmi_origin exists
-	# use that and rename it libmi.
+	# copy the libmi_origin.dylib to libmi.dylib 
+	cp $(OUTPUTDIR)/lib/libmi_origin.dylib $(PACKAGE_DIR)/libmi.dylib
+else #($(PF),Darwin)
```
jeffaco: Prefer a full comment after the 'else'. In this case, `ifeq`. `ifeq ($(PF),Darwin)` is a lot different than `ifneq ($(PF),Darwin)`.

Unix/installbuilder/GNUmakefile
paulcallen: yes, cosmetic

Unix/installbuilder/GNUmakefile
paulcallen: these were very close to each other and so that is why I chose not to, but I am happy to add it

PR #137, opened 2016-12-02
privateKeyFile bugfix

johnkord:
Bug fix for HTTPS client authentication, which CM needs.

@Microsoft/ostc-devs


johnkord:
Thanks for the review, Jeff. I've changed the commit message accordingly!

PR #142, opened 2016-12-12
Fixes auth failure in omiserver

yakman2020:
We need to include an anonymous credential in gss_accept_sec_context as a verifier cred.

### DIFF
--- a/Unix/http/httpauth.c
+++ b/Unix/http/httpauth.c
@@ -1290,6 +1290,10 @@ void Deauthorize(_In_ Http_SR_SocketData * handler)
 #endif
         handler->pAuthContext = hdl;
     }
+    if (handler->pVerifierCred) {
+        (* _g_gssState.Gss_Release_Cred)(&min_stat, handler->pVerifierCred);
+        handler->pVerifierCred = NULL;
+    }    
 
     handler->httpAuthType = AUTH_METHOD_UNSUPPORTED;
     handler->encryptedTransaction = FALSE;
@@ -1454,6 +1458,7 @@ MI_Boolean IsClientAuthorized(_In_ Http_SR_SocketData * handler)
         {
             gss_cred_id_t verifier_cred_handle = GSS_C_NO_CREDENTIAL;
             gss_OID_set actual_mechs = GSS_C_NO_OID_SET;
+
             /* Get acceptor cred for principal. */
             maj_stat = (*_g_gssState.Gss_Acquire_Cred)(&min_stat, GSS_C_NO_NAME, GSS_C_INDEFINITE, mechset, GSS_C_ACCEPT, &verifier_cred_handle, &actual_mechs, NULL); // Name needs to not be null?
             if (_check_gsserr("gss_acquire_cred(acceptor) ", maj_stat, min_stat))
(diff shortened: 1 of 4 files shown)


yakman2020:
@Microsoft/omi-devs 
@jeffaco

Unix/http/httpauth.c
```diff
@@ -1529,9 +1538,12 @@ MI_Boolean IsClientAuthorized(_In_ Http_SR_SocketData * handler)
 
                 (* _g_gssState.Gss_Delete_Sec_Context)(&min_stat, &context_hdl, NULL);
 
-                handler->pAuthContext = NULL;
+                handler->pAuthContext  = NULL;
```
jeffaco: Why add a second space here, when it's inconsistent from all else here?

Unix/http/httpclientauth.c
```diff
@@ -684,11 +684,12 @@ static char *_BuildInitialGssAuthHeader(_In_ HttpClient_SR_SocketData * self, MI
     // gss_OID_set_desc mechset_krb5 = { 1, &mech_krb5 };
     // gss_OID_set_desc mechset_iakerb = { 1, &mech_iakerb };
     const gss_OID_desc mechset_avail_elems[] = {
+        { 6, "\053\006\001\005\005\002" },
         { 10, "\x2b\x06\x01\x04\x01\x82\x37\x02\x02\x0a" }, // For now we start with ntlm
         // mech_krb5,   Not yet
         // mech_iakerb,  Not yet
     };
-    const gss_OID_set_desc mechset_avail = { /*3, */ 1, (gss_OID) mechset_avail_elems };
+    const gss_OID_set_desc mechset_avail = { 2, /* 1 ,*/ (gss_OID) mechset_avail_elems };
```
jeffaco: Why the `/* `, /*` part? Not certain of this? If certain, remove it?

Unix/http/httpauth.c
```diff
@@ -1529,9 +1538,12 @@ MI_Boolean IsClientAuthorized(_In_ Http_SR_SocketData * handler)
...
                 handler->authFailed = TRUE;
 
+                (* _g_gssState.Gss_Release_Cred)(&min_stat, handler->pVerifierCred);
```
jeffaco: No check if `handler->pVerifierCred` is null or not. How does Gss_Release_cred behave in this case? I also noted that `gss_release_cred` is called without checking the return value. This is done in other cases too, for other gss APIs. Is this a good idea? Even if you just note the error and log it and then continue, having this in the log can provide critical information if things go awry.

Unix/http/httpclientauth.c
jeffaco: This is littered with "magic numbers", like at line 687, which you added. Comments as to what these are would be appreciated!

Unix/http/httpauth.c
yakman2020: I'm checking it as good manners, but there is a trace message just above it.  pVerifierCred essentially couldn't be null here.

Unix/http/httpclientauth.c
yakman2020: good point. Added comments.

PR #145, opened 2016-12-14
George/supported platforms

palladia:
1. Determine platform type so that NTLM tests only run on Mac/Linux platforms.
2. Stability modifications to make tests run more consistently.


palladia:
ostc-devs

Issue #148, opened 2016-12-15
"omicli --auth BASIC" fails silently 


palladia:

palladia:
Fixed by #149

PR #149, opened 2016-12-15
Issue #148

palladia:
omicli --auth BASIC      (it expects 'Basic') 
fails silently.


palladia:
ostc-devs pls review.

PR #151, opened 2016-12-19
Changes to allow compilation with gcc 6

yakman2020:
@Microsoft/omi-devs 

Needed to fix some problems with gcc 6.2 to compile on 16.10

### DIFF
--- a/Unix/buildtool
+++ b/Unix/buildtool
@@ -1,5 +1,6 @@
 #!/bin/sh
 
+
 ##==============================================================================
 ##
 ## Extract leading arguments into $arg1 ... $arg9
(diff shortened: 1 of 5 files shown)


Unix/buildtool
```diff
@@ -1,5 +1,48 @@
...
+    os2=$2
+    
+    osmaj1=${os1%.*}
+    osmin1=${os1#*.}
+    
+    osmaj2=${os2%.*}
+    osmin2=${os2#*.}
+    
+    if [ "$osmaj1" -lt "$osmaj2" ]
+    then
```
jeffaco: Fix formatting to be consistent with rest of file (see line 766 as an example).

Will wait for new iteration since you stated this isn't working properly.

Unix/buildtool
```diff
@@ -224,36 +225,50 @@ if [ -z "$target" ]; then
...
             compiler=GNU
+            distro=`uname -s`
+            distro_version=`uname -v`
             ;;
         ppc*:Linux:*)
             os=LINUX
             arch=PPC
             compiler=GNU
+            distro=`lsb_release -i | awk -F":" '{ print $2 }'`
+            distro_version=`lsb_release -r | awk -F":" '{ print $2 }'`
```
jeffaco: Note that this will give you leading spaces, which you may or may not care about:

```
jeffcof:code> lsb_release -r | awk -F":" '{ print $2 }'
        7.3.1611
jeffcof:code>
```

Unix/buildtool
```diff
@@ -734,11 +782,20 @@ if [ "$arg1" = "cflags" -o "$arg1" = "cxxflags" ]; then
...
-	    # _BSD_SOURCE is deprecated on PPC
-            if test "$platform" = "LINUX_PPC_GNU"; then
-		r="$r -D_DEFAULT_SOURCE"
-	    else
-		r="$r -D_BSD_SOURCE"
+	        # _BSD_SOURCE is deprecated on PPC and gcc 6 and later
+
+            if test "$platform" = "LINUX_PPC_GNU" ; then
+		        r="$r -D_DEFAULT_SOURCE"
+	        else
```
jeffaco: Why different indentation than everything else?

PR #156, opened 2016-12-22
Adding extraHeaders to header string

johnkord:
extraHeaders was previously ignored, and this now appends all extraHeaders onto the HTTP headers that are sent.

This fixes a bug associated with CM's NAA/NTLM.

@Microsoft/ostc-devs


johnkord:
I noticed an issue with tabs/spaces, so pushed up a new version with no code changes except replacing the tabs with spaces.

PR #159, opened 2016-12-27
Able to connect using encrypted transmission to/from linux

yakman2020:
Windows is having a problem. I expect I need to encrypt zero length messages. 
But the change is close and on point for linux->linux.

### DIFF
--- a/Unix/http/httpclient.c
+++ b/Unix/http/httpclient.c
@@ -762,6 +762,10 @@ static Http_CallbackResult _ReadData(
     char* buf;
     size_t buf_size, received;
     MI_Result r;
+#if defined(ENCRYPT_DECRYPT)
+     Page *pOldRecvPage = NULL;
+#endif
+
 
     /* are we in the right state? */
     if (handler->recvingState != RECV_STATE_CONTENT)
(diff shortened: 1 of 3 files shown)


Unix/http/httpclient.c
```diff
@@ -813,6 +817,37 @@ static Http_CallbackResult _ReadData(
         _WriteTraceFile(ID_HTTPCLIENTRECVTRACEFILE, (char*)(handler->recvPage + 1), handler->receivedSize);
     }
 
+
+    if (!handler->ssl && handler->encrypting)
+    {
+
+#if ENCRYPT_DECRYPT
```
jeffaco: Why a compile-time conditional for this?

If the compile-time conditional is disabled, but we're supposed to encrypt, then this if will fall through without returning any kind of status/result. Is this correct?

Unix/http/httpclientauth.c
```diff
@@ -461,7 +461,586 @@ static int EncodePlaceCallback(const char *data, size_t size, void *callbackData
...
+                            original_content_length = atoi(linep);
+                            while (';' != *linep && *linep && linep < linelimit)
+                                linep++;
+                            linep++;
+                        }
+                        else if (Strncasecmp(linep, TYPE_FIELD, TYPE_FIELD_LEN) == 0)
+                        {
+                            linep += TYPE_FIELD_LEN;
+                            original_content_type = linep;
+                            while (';' != *linep && *linep && linep < linelimit)
```
jeffaco: Coding standards calls for braces, even for one-line `if` or `while` type statements.

Unix/http/httpclientauth.c
```diff
@@ -461,7 +461,586 @@ static int EncodePlaceCallback(const char *data, size_t size, void *callbackData
...
+    gss_buffer_desc output_buffer = { 0 };
+    Page *page = NULL;
+    char *scanlimit = NULL;
+    char *scanp = NULL;
+    char *segp = NULL;
+    char *linep = NULL;
+    char *linelimit = NULL;
+    int flags = (int)handler->negoFlags;
+    uint32_t sig_len = 0;
+    //uint32_t sig_flags = 0;
```
jeffaco: What is this for? Remove?

Unix/http/httpclientauth.c
```diff
@@ -461,7 +461,586 @@ static int EncodePlaceCallback(const char *data, size_t size, void *callbackData
...
+
+    static const char APPLICATION_SPNEGO[] = "application/HTTP-SPNEGO-session-encrypted";   // 2do: compare to header protocol
+    static const size_t APPLICATION_SPNEGO_LEN = MI_COUNT(APPLICATION_SPNEGO) - 1;
+
+    static const char MULTIPART_ENCRYPTED[] = "multipart/encrypted";
+    static const size_t MULTIPART_ENCRYPTED_LEN = MI_COUNT(MULTIPART_ENCRYPTED) - 1;
+
+    const char *content_type = NULL;
+    const char *content_len_str = NULL;
+    //char *char_set = NULL;
```
jeffaco: What is this for? Remove?

Unix/http/httpclientauth.c
```diff
@@ -461,7 +461,586 @@ static int EncodePlaceCallback(const char *data, size_t size, void *callbackData
...
+    // We know we have the additional room in the page because the string was in the page already
+    memcpy(buffer_p, original_content_type, strlen(original_content_type) + 1);
+    original_content_type = buffer_p;
 
+    buffer_p += strlen(original_content_type) + 1;  // Include the null
+    memcpy(buffer_p, original_encoding, strlen(original_encoding) + 1);
+
+    (*_g_gssClientState.Gss_Release_Buffer)(&min_stat, &output_buffer);
+
+#if 0
```
jeffaco: `#if 0` - why?

Unix/http/httpclient.c
yakman2020: The conditional is a protection requested by Paul to allow us to deactivate the encrypt code if needed. 
Don't know why it would be needed, but there it is.

Unix/http/httpclientauth.c
yakman2020: removed (will show up in the next pull request)

Unix/http/httpclientauth.c
yakman2020: removed (will show up in the next pull request)

Unix/http/httpclient.c
yakman2020: I would expect that if we decided to deactivate encrypt we would either put the same ifdef around the  selection of "--encryption http"  so that it was not possible to select it from the command line or from HttpClient_NEwConnector2

PR #169, opened 2017-01-04
Fixed httpclient.c auth_header propagation in AUTH_METHOD_BYPASS mode

johnkord:


johnkord:
@Microsoft/ostc-devs

Unix/deprecated/httpclientcxx/httpclientcxx.cpp
```diff
@@ -530,7 +530,7 @@ void IOThread::StartRequestTh(NotifyItem* item)
         c_headers.data = &headers_pointers[0];
     }
 
-    headers_pointers.push_back("Authorization: None");
+    headers_pointers.insert(headers_pointers.begin(), "Authorization: None");
```
jeffaco: Order doesn't matter? Or it does and you want the beginning?

johnkord: It does matter, due to some convoluted reasons and the fact that httpclientcxx.cpp is deprecated and doesn't use the new httpclient interface. Definitely want this in the beginning, so that CM can override it by setting its own Authorization header in the extraHeaders.

PR #173, opened 2017-01-06
Added extra spnego auth tests in test_cli.cpp

yakman2020:
Added tests to exercise encryption on http and unencrypted on http.

### DIFF
--- a/Unix/buildtool
+++ b/Unix/buildtool
@@ -175,6 +175,8 @@ COMMANDS:
         Print the directory containing the OpenSSL library
     ldlibrarypath
         Print the name of the dynamic library path variable, usually LD_LIBRARY_PATH or SHLIB_PATH
+    ntlm_supported
+        1 if ntlm gss spnego is available on the platform. 0 otherwise
 EOF
     exit 0
 fi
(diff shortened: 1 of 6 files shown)


Unix/buildtool
```diff
@@ -1753,6 +1755,108 @@ fi
...
+        exit 1
+    fi
+
+    case "$os" in
+        LINUX*)
+            # supported!
+            case "$distro" in
+                RedHat*)
+                    
+                    distro_major_Version=`echo $distro_version | awk -F'.' '{ print $1}'`
```
niroyb: Looks like `distro_major_Version` and `distro_minor_Version` can be calculated once outside the switch case since it's the same for all platforms.

Unix/tools/auth_tests_setup.sh
```diff
@@ -57,51 +57,6 @@ if [ "x${username}" != "x" -a "x${userpasswd}" != "x" ]; then
...
-            fi
-        fi
-    else
-        echo "Running on Linux"
-        supported=1
-    fi
-fi
-
-export NTLM_SUPPORTED_PLATFORM
+supported=`$basedir/buildtool`
```
niroyb: Is the `ntlm` argument missing for buildtool?

PR #175, opened 2017-01-06
Fixes AIX test problems

yakman2020:
@Microsoft/omi-devs


Unix/tests/cli/test_cli.cpp
```diff
@@ -151,6 +151,32 @@ static int StartServer()
...
+        int pid = -1;
+        int status = PIDFile_Read(&pid);
+
+        if (pid != -1)
+        {
+
+#if defined(aix)
+            sprintf(buf, "sudo kill -TERM %d", pid);
+#elif !defined(solaris) && !defined(linux)
+            sprintf(buf, "killall -TERM %d", pid);
```
jeffaco: If you have a specific PID, what's the benefit of `killall` or `pkill`, which generally kill groups of processes by name? With a specific PID, why not just use `kill` universally?

Unix/tests/cli/test_cli.cpp
```diff
@@ -151,6 +151,32 @@ static int StartServer()
...
+#else
+            sprintf(buf, "pkill -TERM %d", pid);
+#endif
+            system(buf);
+            Sleep_Milliseconds(10);
```
jeffaco: I'm pretty uncomfortable with a sleep like this. First, 50 milliseconds is very quick, and I can virtually guarantee that, at times, some of our slower platforms may not keep up.

For this sort of thing, I like do do something like (pseudo-code):

```
int counter = 0;
<tell process to shut down>
while (counter <= 50 && <process still exists>) {
    Sleep_Milliseconds(100);
    counter++;
}

if <process still exists> {
    log error and perhaps abort or try again with kill -9
}
```

In this way, if the process actually goes away quickly, you end quickly. If OMI takes some time to clean up, that's okay too - with 100 milliseconds sleep and a counter limit of 50, you'll have 5 full seconds to shut down, which should normally work.

We've had a LOT of issues over all our platforms with intermittent failures for tests that sleep. Much better to verify that things worked or wait longer.

Also note that, sometimes (I don't know why), we've observed the inability to kill processes, even with -9, particularly on AIX. We ended up rebooting when that occurred, and after that it wouldn't repro.

Unix/tests/util/util.c
```diff
@@ -401,15 +401,24 @@ MI_Result StartOmiserver()
     //
     {
         char buf[64];
+
+        int pid = -1;
+        int status = PIDFile_Read(&pid);
+
+fprintf(stderr, "pid is %d\n", pid);
```
niroyb: An error message in the else case would be more useful when the pid can't be read.

Unix/tests/cli/test_cli.cpp
```diff
@@ -318,6 +344,28 @@ static int StartServerSudo()
...
+        {
+
+#if defined(aix)
+            sprintf(buf, "sudo kill -TERM %d", pid);
+#elif !defined(solaris) && !defined(linux)
+            sprintf(buf, "killall -TERM %d", pid);
+#else
+            sprintf(buf, "pkill -TERM %d", pid);
+#endif
+            system(buf);
```
niroyb: Should you care about the return value of `system`?

Unix/tests/cli/test_cli.cpp
```diff
@@ -139,6 +139,64 @@ bool ParseCommandLine(MI_Char command[PAL_MAX_PATH_SIZE], vector<MI_Char*>& args
...
+**
+** Kill Old Server Instance
+**
+   ============================================================================*/
+
+/* We don't have a return value because this function will terminate the test 
+ * if it can't kill the previous omiserver. So success or program exit
+ */
+
+void KillOldServerInstance()
```
jeffaco: Rather than having two instances, please try and grab the copy from util.c by linking against that.

Other than that, this looks fine. I'd prefer if it didn't hang for 2.5 seconds before recovering from a `kill -9` operation, but I understand your reluctance to change product code unnecessarily.

PR #178, opened 2017-01-09
George/delete pidfile

palladia:
Instead of a sleep, we forcibly delete the pid file, and wait to make sure all children processes are done.


palladia:

PR #179, opened 2017-01-10
Add SessionId support to wsman client

paulcallen:
SessionId soap header is needed as part of PSRP shell protocol to enable
connect/disconnect logic.

@Microsoft/omi-devs

### DIFF
--- a/Unix/wsman/wsbuf.c
+++ b/Unix/wsman/wsbuf.c
@@ -3205,11 +3205,20 @@ static MI_Result WSBuf_CreateRequestHeader(WSBuf *buf,
     }
 
 #ifndef DISABLE_SHELL
-   if (cliHeaders->compressionType)
+    if (cliHeaders->compressionType)
     {
         if (MI_RESULT_OK != WSBuf_AddLit(buf, LIT(ZT("<h:CompressionType s:mustUnderstand=\"true\">"))) ||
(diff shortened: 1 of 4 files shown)


Unix/wsman/wsbuf.c
```diff
@@ -3214,6 +3214,15 @@ static MI_Result WSBuf_CreateRequestHeader(WSBuf *buf,
             goto failed;
         }
     }
+    if (cliHeaders->sessionId)
+    {
+        if (MI_RESULT_OK != WSBuf_AddLit(buf, LIT(ZT("<p:SessionId s:mustUnderstand=\"false\">"))) ||
+            MI_RESULT_OK != WSBuf_AddStringNoEncoding(buf, cliHeaders->sessionId) ||
+            MI_RESULT_OK != WSBuf_AddLit(buf, LIT(ZT("</p:SessionId>"))))
+        {
+            goto failed;
```
jeffaco: There are three different reasons that `goto failed` will get control. Do we ever care which of these failures cause the problem?

I see it's a common calling construct. I'm just wondering if, long term, that's a good approach. It depends on when/how often failures can occur ...

paulcallen: they all can fail due to out of memory, so no it does not matter.

PR #183, opened 2017-01-11
Fix Nego auth issue

yakman2020:
@Microsoft/omi-devs

### DIFF
--- a/Unix/http/httpauth.c
+++ b/Unix/http/httpauth.c
@@ -1689,16 +1689,17 @@ MI_Boolean IsClientAuthorized(_In_ Http_SR_SocketData * handler)
 #ifdef AUTHORIZATION
     else
     {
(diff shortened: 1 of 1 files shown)


Unix/http/httpauth.c
```diff
@@ -1734,7 +1735,7 @@ MI_Boolean IsClientAuthorized(_In_ Http_SR_SocketData * handler)
 #endif      
 
             protocol_p = AUTHENTICATION_NEGOTIATE;
-            mechset = (gss_OID_set) & mechset_spnego;
+            mechset = (gss_OID_set) & mechset_avail;
```
paulcallen: any time there is a cast I get cautious. Is this really a safe cast?

yakman2020: yes. The set is a pointer.

PR #185, opened 2017-01-12
George/warnings

palladia:
1.  Treat compiler warnings as errors on HPUX and AIX.
2.  Treat compiler warnings as errors on last part of regress (output4 builds).
3.  Fix code to not generate warnings.

### DIFF
--- a/Unix/omi_error/omierror.c
+++ b/Unix/omi_error/omierror.c
@@ -382,12 +382,15 @@ MI_INLINE const MI_Char *Errno_ToString(
     }
 # else /* defined(CONFIG_ENABLE_WCHAR) */
     {
-        *buffer = '\0';
-
-        // Intentionally disregarding return code
-        (void)strerror_r(OMI_Code, buffer, len) ;
-
+#if defined(macos) || ((_POSIX_C_SOURCE >= 200112L || _XOPEN_SOURCE >= 600) && ! _GNU_SOURCE)
+        int ret = strerror_r(OMI_Code, buffer, len);
+        if (ret != 0)
+            *buffer = '\0';
         return buffer;
+#else
+        char *ret = strerror_r(OMI_Code, buffer, len);
+        return ret;
+#endif
     }
 # endif /* defined(CONFIG_ENABLE_WCHAR) */
 }
--- a/Unix/regress
+++ b/Unix/regress
@@ -278,7 +278,7 @@ fi
 ##
 
 rm -rf $prefix
-./configure --prefix=$prefix --outputdirname=$current_output $options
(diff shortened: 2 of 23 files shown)


palladia:

Unix/omi_error/omierror.c
```diff
@@ -382,12 +382,15 @@ MI_INLINE const MI_Char *Errno_ToString(
...
-        // Intentionally disregarding return code
-        (void)strerror_r(OMI_Code, buffer, len) ;
-
+#if defined(macos) || ((_POSIX_C_SOURCE >= 200112L || _XOPEN_SOURCE >= 600) && ! _GNU_SOURCE)
+        int ret = strerror_r(OMI_Code, buffer, len);
+        if (ret != 0)
+            *buffer = '\0';
         return buffer;
+#else
+        char *ret = strerror_r(OMI_Code, buffer, len);
```
jeffaco: I'm not familiar with POSIX rules in specific versions.

I'm unsure if this is correct. I guess, if an error is returned, you have no way to return that, and that's why you're zeroing the buffer (386-389)? But then at 391-392, that's not the case. Seems like an inconsistency in your behavior.

Is this really what you want?

Unix/regress
```diff
@@ -278,7 +278,7 @@ fi
 ##
 
 rm -rf $prefix
-./configure --prefix=$prefix --outputdirname=$current_output $options
+./configure --enable-werror --prefix=$prefix --outputdirname=$current_output $options
```
jeffaco: Do we want --enable-werror to be the default if --enable-microsoft?

Unix/omi_error/omierror.c
palladia: According to manual pages, if strerror_r returns char* (the so called GNU version), if an error occurs, the returned string would be "Unknown error nnn".

Unix/regress
palladia: --enable-werror in this case catches some compilation problems you don't get with --dev, since it's not a DEBUG build.

If --enable-microsoft is for run-time only, it's probably not necessary to also enable --enable-werror.

Unix/omi_error/omierror.c
jeffaco: Wow, that's a really stupid API interface. But it is what it is!

Unix/regress
jeffaco: --enable-microsoft is how the nightly builds will build OMI. So the question is, should the nightly builds use --enable-werror, which would guarantee no regressions in this area?

paulcallen: I see no harm in enabling it for Microsoft builds. We have fixed the necessary issues now, so lets keep it clean.

Issue #184, opened 2017-01-11
Basic auth using incomplete username if 'user@domain' is used


paulcallen:
only user is encoded in the auth header rather than full user@domain.
basic auth should use the username as is with no changes.

yakman2020:
Fixed by commit 86eea9e793f9b30a5ae099a8f69fa138edfcd09b

PR #190, opened 2017-01-13
Alow complex domain credentials in basic auth. Fixes Issue #184

yakman2020:


Unix/http/httpclient.c
```diff
@@ -2566,32 +2566,43 @@ MI_Result HttpClient_New_Connector2(
...
+                self->connector->username    = username;
+                self->connector->user_domain = NULL;
             }
-            else
+            else if ( authtype == AUTH_METHOD_NEGOTIATE ||
+                      authtype == AUTH_METHOD_NEGOTIATE_WITH_CREDS ||
+                      authtype == AUTH_METHOD_KERBEROS) 
             {
-                user_domain = memchr(username, '@', strlen(username));
+                user_domain = memchr(username, '\\', strlen(username));
```
paulcallen: strchr is probably quicker as your solution always searches for the null terminator where strchr looks for the char until it hits the null terminator.

Unix/http/httpclient.c
```diff
@@ -2566,32 +2566,43 @@ MI_Result HttpClient_New_Connector2(
...
-                    self->connector->user_domain = PAL_Strdup(user_domain+1);
+                    self->connector->username    = PAL_Strdup(user_domain+1);
+                    self->connector->user_domain = username;
                 }
                 else
                 {
-                    // No credential domain specified
-                    self->connector->username    = username;
-                    self->connector->user_domain = PAL_Strdup((char*)host);
+                    user_domain = memchr(username, '@', strlen(username));
```
paulcallen: same comment about strchr

Unix/tests/http/test_httpclient.cpp
```diff
@@ -141,6 +146,33 @@ void _StrandTestClientPost( _In_ Strand* self_, _In_ Message* msg )
     else
         s_charset = "";
 
+    if (request->headers->authorization)
```
paulcallen: aren't you going to want to crack open the basic auth header to validate it is encoded properly in there? That would be more complete.

paulcallen:
can you add the issue number to the description so they are tied together.

Unix/tests/http/test_httpclient.cpp
yakman2020: _StrandTestClientPost has a request which is coming from the server, after parsing of the http headers. I don't have access to the original auth header. But the password and username members of the HttpRequestMsg struct which I do get are busted and parsed exactly as the server is seeing them, so the work (which I actually built but couldn't use) wasn't necessary.

Of course that is only true about Basic.  For Nego we would have to figure something else out, but this isn't about Nego today

Unix/http/httpclient.c
yakman2020: done

Unix/http/httpclient.c
yakman2020: done

Unix/tests/http/test_httpclient.cpp
paulcallen: ok, my bad

paulcallen:
fixes issue #184

Issue #194, opened 2017-01-16
omicli segvs with "not supported" on auth failure


yakman2020:
The reason is the failure to end the strand. If an enumerate was started an enuerate context is still active and the EnumeratePullRequest tries to inspect the not-existing headers. 

The solution is to deactivate the enumerate context and post a bad result to the strand.

PR #195, opened 2017-01-16
omicli should gracefully fail the request if auth fails 

yakman2020:
Fixes issue #194 
@Microsoft/omi-devs

### DIFF
--- a/Unix/wsman/wsmanclient.c
+++ b/Unix/wsman/wsmanclient.c
@@ -71,6 +71,7 @@ struct _WsmanClient
     WSBuf wsbuf;
     MI_Uint32 httpError;
     EnumerationState *enumerationState;
+    ptrdiff_t receivedPost;
     ptrdiff_t ackOriginalPost;
     MI_Uint32 maxElements;
     Page *responsePage;
@@ -680,6 +681,8 @@ void _WsmanClient_Post( _In_ Strand* self_, _In_ Message* msg)
 
     DEBUG_ASSERT( NULL != self_ );
 
+    self->receivedPost = MI_TRUE;
+
 //    DEBUG_ASSERT(self->message == NULL);
 //    self->base.mask &= ~SELECTOR_READ;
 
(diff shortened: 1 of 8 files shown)


Unix/etc/omicli.conf
```diff
@@ -38,6 +38,6 @@
 ##
 ## loglevel,logpath,logfile -- set the loggiing options for MI client
 ##
-#loglevel = DEBUG
+loglevel = VERBOSE
 #logpath = var/log/
-#logfile = miclient.log
\ No newline at end of file
+#logfile = miclient.log
```
jeffaco: No longer a unix-format file, please fix.

Unix/wsman/wsmanclient.c
```diff
@@ -143,12 +145,23 @@ static void HttpClientCallbackOnStatusFn2(
         const ZChar *text)
 {
     WsmanClient *self = (WsmanClient*) callbackData;
-    if (!self->enumerationState || self->enumerationState->endOfSequence)
+
+    
+    if (result == MI_RESULT_TIME_OUT)
```
jeffaco: Brief comment on why you're translating to MI_RESULT_INVALID_PARAMETER

jeffaco:
Please insure that the CI test failures on Linux are not due to this, and help figure out why regress is failing on CI. CI used to be the ONLY place regress worked!

Unix/etc/omicli.conf
paulcallen: Should not check in this change

Unix/etc/omiserver.conf
```diff
@@ -23,7 +23,7 @@
 ##
 ## loglevel -- set the log level of the server
 ##
-loglevel = DEBUG
+loglevel = VERBOSE
```
paulcallen: This should be either disabled altogether, or WARNING should be specified in the checked in version

Unix/wsman/wsmanclient.c
```diff
@@ -143,12 +145,23 @@ static void HttpClientCallbackOnStatusFn2(
...
+        result = MI_RESULT_INVALID_PARAMETER;
```
paulcallen: should be MI_RESULT_FAILED WITH an OMI_Error object, and set probable cause to 111 - timeout.

Issue #200, opened 2017-01-19
#pragma once not supported on all platforms


palladia:
This causes compilation errors on AIX.

palladia:
Fixed by #201

PR #201, opened 2017-01-19
Issue #200: Replace #pragma once on wsmanerrorhandling.h

palladia:


palladia:
@Microsoft/ostc-devs

Unix/wsman/wsmanerrorhandling.h
```diff
@@ -308,5 +309,4 @@ const Probable_Cause_Data* GetWsmanCimError(Error_Types type);
 
 END_EXTERNC
 
-
-
+#endif
```
jeffaco: By convention, since this `#endif` is hundreds of lines later, I prefer something like:

```#endif // !defined(_wsmanerrorhandling_h)```

This makes it easier to see what's what in the header file.

palladia: Fixed.

PR #202, opened 2017-01-20
Is shell

paulcallen:
@Microsoft/omi-devs @yakman2020 @palladia 

Update client-side wsman processing such that the packets deal with shell specific functionality only if told to do so with an IsShell option.


jeffaco:
First, please note that your commit message is too long. I get:

```
    use IsShell destination option instead of resourceuri
    
    resourceuri can change for shells so need more reliable way to determine if it is a shell on client
```

Line wrap before "determine" ...

Issue #196, opened 2017-01-18
omicli -xml crashes when response includes certain characters


gorankor:
omicli always crashes on this request for me:

$ omicli -t **-xml** -u ... -p ... --hostname ... gi root/cimv2 '{' Win32_Service name spooler '}'

The crash happens here when accessing the "_encode" array outside of its bounds:

Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000000000040fd3b in WriteBuffer_Char (clientBuffer=0x1df59f0 "<INSTANCE CLASSNAME=\"Win32_Service\"><PROPERTY NAME=\"AcceptPause\" TYPE=\"string\"><VALUE>false</VALUE></PROPERTY><PROPERTY NAME=\"AcceptStop\" TYPE=\"string\"><VALUE>true</VALUE></PROPERTY><PROPERTY NAME=\"Ca"..., clientBufferLength=10000, clientBufferNeeded=0x7ffccb193ecc, charToWrite=-30 '\342', escapingDepth=1, result=0x7ffccb193e18) at xmlserializer.c:894
894             (_encode[(unsigned int)charToWrite].str == NULL))

Here's the interesting part of the received string - the problematic character being E2:

$ tail -1 omiclient-recv.trc  | xxd
...
0000490: 7365 7276 6963 652c 2079 6f75 2077 6f6e  service, you won
00004a0: **e2**80 9974 2062 6520 6162 6c65 2074 6f20  ...t be able to
00004b0: 7072 696e 7420 6f72 2073 6565 2079 6f75  print or see you
00004c0: 7220 7072 696e 7465 7273 2e3c 2f70 3a44  r printers.</p:D
...

The fix is simple - just ensure array index is OK before accessing it.

PR #206, opened 2017-01-21
XML handling of chars outside normal range fail

paulcallen:
xml.c was a hang, xmlserializer.c was a crash.
Fixes issue #196

@Microsoft/ostc-devs

### DIFF
--- a/Unix/xmlserializer/xmlserializer.c
+++ b/Unix/xmlserializer/xmlserializer.c
@@ -890,7 +890,7 @@ static void WriteBuffer_Char(
     };
 
     if ((escapingDepth == 0) ||
-        (charToWrite > L'>') ||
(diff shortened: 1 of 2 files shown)


Unix/xmlserializer/xmlserializer.c
```diff
@@ -890,7 +890,7 @@ static void WriteBuffer_Char(
     };
 
     if ((escapingDepth == 0) ||
-        (charToWrite > L'>') ||
+        (((unsigned int)charToWrite) > L'>') ||
```
niroyb: The length of 'unsigned int' is two or 4 bytes. Consider using wchar_t instead. It is also of variable lenght but it will be the same size as `L'<'`

PR #211, opened 2017-01-26
Call pam_setcred to corectly work with pam_tally2

yakman2020:
This fixes the compat issue with pam-tally2 accumulating "bad" login attempts.

### DIFF
--- a/Unix/base/user.c
+++ b/Unix/base/user.c
@@ -225,6 +225,12 @@ static int _PamCheckUser(
         return -1;
     }
 
+    if (PAM_SUCCESS != pam_setcred(t, PAM_ESTABLISH_CRED))
+    {
+        pam_end(t,0);
+        return -1;
+    }
+
     pam_end(t, 0);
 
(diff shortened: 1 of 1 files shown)


Unix/base/user.c
```diff
@@ -225,6 +225,12 @@ static int _PamCheckUser(
         return -1;
     }
 
+    if (PAM_SUCCESS != pam_setcred(t, PAM_ESTABLISH_CRED))
+    {
+        pam_end(t,0);
+        return -1;
+    }
+
```
jeffaco: Given this has been PBUILT already, it's hardly worth it. But you could have replaced 228-236 with:

```
int retCode = ( PAM_SUCCESS != pam_setcred(t, PAM_ESTABLISH_CRED) ? -1 : 0 );
pam_end(t, 0);
return retCode;
```

This is just an FYI, I don't recommend this change (not worth the retesting effort) unless you feel strongly.

Issue #212, opened 2017-01-30
Chunking tests fail occassionally


palladia:

palladia:
Fixed by #213

PR #213, opened 2017-01-30
Issue #212: Repeat read if EAGAIN is detected

palladia:
Extract chunking changes from paul/george, dealing only on code that caused chunking tests to occassionally fail.  Enabling debug prints show that the socket reads fail with EAGAIN.


palladia:

PR #215, opened 2017-01-31
Remove brew install openssl

palladia:
openssl is now installed by default.  Forcing it to install again causes errors in builds.


palladia:
@Microsoft/ostc-devs

PR #216, opened 2017-02-01
Create NTLM cred file control in omiserver

yakman2020:
@Microsoft/omi-devs 

Elminates the need  for  the NTLM_USER_FILE env var.
Ignores NTLM_USER_FILE if set

### DIFF
--- a/Unix/base/user.c
+++ b/Unix/base/user.c
@@ -138,7 +138,7 @@ const char* GetHomeDir()
     }
 
     /* copy pw_dir since we do not own pwd */
-    home = strdup(pwd->pw_dir);
+    home = PAL_Strdup(pwd->pw_dir);
 
     return (const char*)home;
 }
@@ -718,6 +718,84 @@ int FormatLogFileName(
 
     return 0;
 }
+
+
+MI_Boolean ValidateNtlmCredsFile(const char *credFilePath)
+
+{
+    char *cred_dir = PAL_Strdup(credFilePath);
+
+    if (!cred_dir)
+    {
+        // Log error?
+        return MI_FALSE;
+    }
+
+    char* p = strrchr(cred_dir, '/');
+    if (p)
+    {
+        *p = '\0';
+    }
+
+    {
+        struct stat buf = {0};
+        int rtn = stat(credFilePath, &buf);
+
+        if (rtn < 0)
+        {
+            // The file not existing is a failure. We said it was there or we wouldn't be here
+            //
+            goto Err;
+        }
+
+        if (S_ISDIR(buf.st_mode) || S_ISLNK(buf.st_mode))
+        {
+            // Not a file? Unlikely, but complain and issue error
+            goto Err;
+        }
+
+        // Acceptable dir will be user only, no others
+        if (!(buf.st_mode & S_IRUSR) || ( buf.st_mode & (S_IRWXG|S_IRWXO)))
+        {
+            goto Err;
+        }
+    }
+
+    // Check the permissions on the directory. They need to be read-only owner
+        
+    {
+        struct stat buf = {0};
+
+        int rtn = stat(cred_dir, &buf);
+        if (rtn < 0)
+        {
+            goto Err;
+        }
+
+        if (!S_ISDIR(buf.st_mode))
+        {
+           // Not a directory? Unlikely, but complain and issue error
+            goto Err;
+        }
+
+        // Acceptable dir will be user only, no others
+        if (!(buf.st_mode & S_IRUSR) || ( buf.st_mode & (S_IRWXG|S_IRWXO)))
+        {
+            goto Err;
+        }
+    }        
+
+    setenv("NTLM_USER_FILE", credFilePath, 1);
+
+    PAL_Free(cred_dir);
+    return TRUE;
+
+Err:
+    PAL_Free(cred_dir);
+    return FALSE;
+
+}
+
 
 #endif
 
--- a/Unix/configure
+++ b/Unix/configure
@@ -74,6 +74,7 @@ set_microsoft_build_options()
     localstatedir=/var/opt/omi
     sysconfdir=/etc/opt/omi/conf
     certsdir=/etc/opt/omi/ssl
+    credsdir=/etc/opt/omi/creds
     enable_native_kits=1
 }
 
@@ -530,7 +531,7 @@ if [ "$dev" = "1" ]; then
   bindir=$outputdir/bin
   localstatedir=$outputdir/var
   includedir=$outputdir/include
-  sysconfdir=$root/etc
+  sysconfdir=$outputdir/etc
 fi
 
 ##==============================================================================
@@ -622,6 +623,10 @@ fi
 
 if [ -z "$certsdir" ]; then
     certsdir=$sysconfdir/ssl/certs
+fi
+
+if [ -z "$credsdir" ]; then
+    credsdir=$sysconfdir/creds
 fi
 
 if [ -z "$authdir" ]; then
@@ -727,6 +732,7 @@ mkdir -p $outputdir/var/log
 mkdir -p $outputdir/var/run
 mkdir -p $outputdir/var/omiauth
 mkdir -p $outputdir/include
+mkdir -p $outputdir/etc
 mkdir -p $tmpdir
 
 ##==============================================================================
@@ -1588,6 +1594,7 @@ CONFIG_LOCALSTATEDIR=$localstatedir
 CONFIG_SYSCONFDIR=$sysconfdir
 CONFIG_PROVIDERDIR=$providerdir
 CONFIG_CERTSDIR=$certsdir
+CONFIG_CREDSDIR=$credsdir
 CONFIG_AUTHDIR=$authdir
 CONFIG_NAMESPACE=$namespace
 CONFIG_HTTPPORT=$httpport
@@ -1622,6 +1629,38 @@ EOF
 
 echo "created $fn"
 
+if [ "$dev" = "1" ]; then
+    ##==============================================================================
+    ##
+    ## Copy and modify etc directory contents as needed for dev
+    ##
+    ##==============================================================================
+
+    fn=$outputdir/etc
+
+    sudo cp -p $rootrelative/etc/omiserver.conf $fn
+    sudo cp -p $rootrelative/etc/omicli.conf $fn
+    sudo cp -p $rootrelative/etc/omigen.conf $fn
+    sudo cp -rp $rootrelative/etc/omiregister $fn
+
+    if [ -f $fn/omiserver.conf ]; then
+      echo "created $fn/omiserver.conf"
+      mkdir -p $credsdir
+      cp -rp $rootrelative/etc/creds/* $credsdir
+      sudo chmod 700 $credsdir
+      sudo chown -R root:root $credsdir
+      test -d $credsdir && echo "created $credsdir"
+      
+      echo "NtlmCredsFile=$credsdir/ntlm" >>$fn/omiserver.conf # register the creds file
+    else
+      echo "required file $fn/omiserver.conf not found"
+      exit -1
+    fi
+
+    test -f $fn/omicli.conf && echo "created $fn/omicli.conf"
+    test -f $fn/omigen.conf && echo "created $fn/omigen.conf"
+    test -d $fn/omiregister && echo "created $fn/omiregister"
+fi
 
 ##==============================================================================
 ##
@@ -1647,6 +1686,10 @@ fi
 chmod +w $fnnitsargs
 
 ntlm_supported=`$buildtool ntlm`
+if [ $ntlm_supported ]; then 
+   echo "NTLM auth can be supported on this platform"
+fi
+
 echo "created $fnnitsargs"
 
 ##==============================================================================
@@ -1690,6 +1733,7 @@ cat > $fn <<EOF
 #define CONFIG_SYSCONFDIR "$sysconfdir"
 #define CONFIG_PROVIDERDIR "$providerdir"
 #define CONFIG_CERTSDIR "$certsdir"
+#define CONFIG_CREDSDIR "$credsdir"
 #define CONFIG_AUTHDIR "$authdir"
 #define CONFIG_NAMESPACE "$namespace"
 #define CONFIG_HTTPPORT $httpport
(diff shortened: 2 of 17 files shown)


Unix/base/paths.c
```diff
@@ -36,6 +36,7 @@
 #define HTTPRECVTRACEFILE LOGDIR "/omiserver-recv.trc"
 #define HTTPCLIENTSENDTRACEFILE LOGDIR "/omiclient-send.trc"
 #define HTTPCLIENTRECVTRACEFILE LOGDIR "/omiclient-recv.trc"
+#define NTLMCREDSFILE CONFIG_CREDSDIR "/ntlm"
```
jeffaco: According to Paul, he would like the location of this file read from omiserver.conf. Talk to him and bring closure to this issue please.

Unix/configure
```diff
@@ -154,6 +155,9 @@ do
       certsdir=$arg 
       ;;
 
+    --credsdir=*)
+      credsdir=$arg 
+      ;;
```
jeffaco: I think this will go away if read from the server. But if it stays, all other stuff seems to have a blank line after `;;`, you should after this section too.

Unix/doc/setup-ntlm-omi.md
```diff
@@ -180,6 +180,17 @@ On the command line:
 
 The most secure method for providing omiserver credentials is via winbind. There may be reasons why this would not be 
 practical for a particular implementation, in which case it is possible to define a credential file, using the same format as
-described for the client.  If this method is used, the location of the file is specified in the omiserver.conf file.
+described for the client.  If this method is used, use of the file is specified by setting the option useNtlmCredsFile=true.
+The file is kept in a standard place ( /etc/opt/omi/creds/ntlm in the release version, output/etc/creds for configure --dev )
```
jeffaco: Nice too keep the docs up to date, good job. That said, I think this is wrong ... (needs to be read in from omiserver.conf).

Unix/etc/omiserver.conf
```diff
@@ -51,3 +51,4 @@ loglevel = DEBUG
 #configfile=PATH
 NoSSLv2=true
 #NoSSLv3=false
+useNtlmCredsFile=true
```
jeffaco: Hmm, there is already a setting here. Seems like you can just decide if you should use it if it's defined or not ...

Unix/base/user.c
```diff
@@ -138,7 +138,7 @@ const char* GetHomeDir()
     }
 
     /* copy pw_dir since we do not own pwd */
-    home = strdup(pwd->pw_dir);
+    home = PAL_Strdup(pwd->pw_dir);
```
paulcallen: this will require a change in the PSL codebase too

Unix/configure
jeffaco: If the NTLM file is coming out of the omiserver.conf file, then why does this flag still exist, and what does it control?

Also, my prior comment was disregarded. Please fix formatting if this does belong (and explain why so I understand). Still looking for a blank line after 160 like everything around it has. Code should be consistently formatted.

Unix/configure
```diff
@@ -2025,10 +2073,16 @@ echo "created $fn"
 ##==============================================================================
 
 if [ "$dev" = "1" ]; then
+  mkdir -p $certsdir
+  echo "built certs dir " $certsdir 
   openssl req -x509 -sha1 -newkey rsa:2048 -days 3650 -nodes -config $fn -keyout $certsdir/omikey.pem -out $certsdir/omi.pem
   openssl x509 -outform der -in $certsdir/omi.pem -out $certsdir/omi.der
   chmod 600 $certsdir/omikey.pem
   chmod 644 $certsdir/omi.pem
+  echo "built certs in " $certsdir
```
jeffaco: Is this a duplicate of line 2077? Or should 2077 say "Building certs in dir ..." ?

Unix/doc/setup-ntlm-omi.md
```diff
@@ -177,9 +177,92 @@ On the command line:
...
-described for the client.  If this method is used, the location of the file is specified in the omiserver.conf file.
+If this method is used, use of the file is specified by setting the option 
+```
+NtlmCredsFile= *path* 
+```
+in the configuration file omiserver.conf.  The file then refered to by the server for credentials. Like certificate directories,
+the server requires permissions of 700 for the directory and 600 for the file. Both the file and directory must be owned by root.
+If the correct permissions and ownership are not set the file will be ignored with an error in the server log.  
+
+The creds/ntlm file must contain credentials of each authorised user with each domain (including hostname). IP addresses
```
jeffaco: Spelling: authorized

Unix/doc/setup-ntlm-omi.md
```diff
@@ -177,9 +177,92 @@ On the command line:
...
+The creds/ntlm file must contain credentials of each authorised user with each domain (including hostname). IP addresses 
+in the NTLM domain are treated as different from hostname, so a separate entry is required for each.
+
+
+## Troubleshooting Authentication
+
+
+###The Process:
+
+-the omi client requests authentication using the username and optional password to acquire a stored credential.
```
jeffaco: These aren't displaying properly in markdown, likely because of lack of space after the dash. Please fix and view the file in the PR before posting again. I'd like folks to review their own markdown themselves prior to posting a PR since it's easy to make a mistake in this area.

Unix/doc/setup-ntlm-omi.md
```diff
@@ -177,9 +177,92 @@ On the command line:
...
+Since it is the most secure, by default, only NTLM authentication via winbind is configured in omiserver. There may be reasons
+why this would not be desirable for a particular implementation, in which case it is possible to define a credential file,
+using the same format as described for the client. 
 
-The most secure method for providing omiserver credentials is via winbind. There may be reasons why this would not be 
-practical for a particular implementation, in which case it is possible to define a credential file, using the same format as
-described for the client.  If this method is used, the location of the file is specified in the omiserver.conf file.
+If this method is used, use of the file is specified by setting the option 
+```
+NtlmCredsFile= *path*
```
jeffaco: Since this is effectively documentation, please make the above line the **suggested** entry, and not what the server would normally accept.

I think, in general, we recommend `tag=value` without a space.

BTW, since this isn't quoted, would spaces be accepted in the filename? You should test that and, if not, say something here about that (or fix it so they are, but that can be tricky without accepting quote characters in case of leading/trailing spaces).

Unix/server/server.c
```diff
@@ -894,6 +895,13 @@ static void GetConfigFileOptions()
             if (SetPathFromNickname(key, value) != 0)
                 err(ZT("SetPathFromNickname() failed"));
         }
+        else if (strcasecmp(key, "NtlmCredsFile") == 0)
+        {
+            if (value)
+            {
+                s_opts.ntlmCredFile = PAL_Strdup(value);
```
jeffaco: So, as I said above, what happens if there are spaces in the filename? Or what if there is a trailing space here on the line? I don't see code to get rid of leading/trailing spaces (or quote characters). If I'm correct, then you either need to improve handling here for parsing or be quite explicit in your documentation.

It would be easy for a trailing space to mess things up, and not be clear why NTLM is not working properly.

Unix/server/server.c
```diff
@@ -1047,6 +1055,25 @@ int servermain(int argc, const char* argv[])
 
 #endif
 
+    /* If ntlm cred file is in use, check permissions and set NTLM_USER_FILE env variable */
+
+    char *ntlm_user_file = getenv("NTLM_USER_FILE");
```
jeffaco: This case is easy since shell has much for flexible parsing for environment variables ...

Unix/tools/auth_tests_cleanup.sh
```diff
@@ -2,12 +2,12 @@
...
+#ntlm_file=$1/etc/ntlm
 
-if [ -f $ntlm_file ]; then
-    echo "Removing NTLM file: $ntlm_file"
-    rm -f $ntlm_file
-fi
+#if [ -f $ntlm_file ]; then
+#    echo "Removing NTLM file: $ntlm_file"
+#    sudo rm -f $ntlm_file
+#fi
```
jeffaco: So why is this block commented out? Should it just be removed entirely? Sometimes it's useful to check in a commented-out block, but often it's just because we forgot to do the "right thing" (fix it or get rid of it).

Unix/tools/auth_tests_setup.sh
```diff
@@ -37,34 +40,60 @@ do
 done
 
 #
-# if required env variables don't exist, then quit
+# Setup the NTLM auth file for client and server. They must be different files because 
+# if the differeing permissions, as the server file must be owned by root and the client by
```
jeffaco: Out of curiosity, must the file be owned by root? Do you validate that? What if we move to a non-prived OMI server? Is it acceptable to be owned by the OMI server account, in that case, or would we need a code change?

Unix/tools/auth_tests_setup.sh
```diff
@@ -37,34 +40,60 @@ do
...
+# the user.
+#
+# If there is already a cred file present we defer to it. In configure we possibly copied over
+# a prebuilt cred file for the server from Unix/etc/creds. If such a file exists, it must contain the test
+# credentials.
```
jeffaco: I'm not clear on the above. Any "configure"d omiserver.conf would never be copied over for the user, unless of special handing due to --dev or something. Make that clear or this can be very confusing since you're referring to something that is "non-standard" (most users never build, much less build with --dev ... only we do that).

jeffaco: Oh, this is in test code. Test code requires --dev. Just make that clear here by stating it.

Unix/tools/auth_tests_setup.sh
```diff
@@ -37,34 +40,60 @@ do
...
+    # configure already created the dir
+
+    server_ntlm_file=$outputdir/etc/creds/ntlm
+    if sudo test ! -f $server_ntlm_file ; then
+       echo "build server auth file " $server_ntlm_file
+       sudo cp $client_ntlm_file $server_ntlm_file
+    else 
+       echo "do not build server auth file " $server_ntlm_file
+    fi
+    sudo chown root:root $server_ntlm_file
```
jeffaco: Where is the "root:root" requirement coming in? If code is validating that, it should be mentioned in my spec. Please give me a filename/filenames and line numbers for this, thanks.

Unix/tools/auth_tests_setup.sh
yakman2020: For now, the file must be owned by the effective user.  If we go to  a non-root user, then it should be owned by that user. Who else would we want it owned by?

jeffaco: If we made the non-root changes, then we'd create an account (like "omiserver") on install, and files would be owned by that.

Give me a brief list of what to change if we do this work, and I'll put it in the spec. That will allow closure of this specific issue.

Unix/configure
yakman2020: I removed the lines. As you mention, its validity is limited to --dev, albeit I would not want it simply to be blank.

Unix/base/user.c
yakman2020: I need permission to check in.

Unix/tools/auth_tests_setup.sh
yakman2020: I added additional text to that effect. Frankly to my eyes it looked pretty clear already, but suspect that that is exactly why we have reviews.

Unix/server/server.c
yakman2020: The value is pre-parsed according to the same rules as every other value in the file.

I'm not clear what case we are trying to solve.

Unix/server/server.c
yakman2020: Again, I don't understand an objection here. The environment variable is whatever the environment variable gets set as. bash does the parsing.

Unix/tools/auth_tests_setup.sh
yakman2020: It shouldn't require any change in the product code. In  ValidateNtlmCredsFile the check is made to match the current uid/gid. If the server is running as root, then root, if it is running as omi_server_daemon, then that would be the ownership required.  

The file auth_tests_setup.sh would require changing one line (line 97) .

I don't beleive there would be any other changes.

Unix/tools/auth_tests_setup.sh
yakman2020: See above. base/user.c ValidateNtlmCredsFile checks for ownership by the uid/gid of the server.  If the server starts running as non root, that line would have to change to the user id/group id the server is using.

Unix/tools/auth_tests_cleanup.sh
yakman2020: I deleted the lines.  I was concerned there would be some complication requiring the ntlm creds file to be destroyed but I don't see any.   The only "vulnerability" would be on a dev machine.

Unix/server/server.c
jeffaco: It looks like you fixed the documentation. That resolves this issue, I think as documented, this would not have worked.

jeffaco:
You're missing a file that should be modified. Please see omi/Unix/installbuilder/conf/omiserver.conf.

Like with `httpport` and `httpsport`, I'd like you to edit the sample omiserver.conf file (that gets installed with the kit) to include comments about what to enter if you want NTLM support in the server.

Please add that to your branch, push an update, and let me know. I'll take a look at the final product and sign off.

Unix/configure
yakman2020: modified

yakman2020:
@jeffaco  modified iaw feedback

Unix/doc/setup-ntlm-omi.md
```diff
@@ -177,9 +177,89 @@ On the command line:
...
+The creds/ntlm file must contain credentials of each authorized user with each domain (including hostname). IP addresses 
+in the NTLM domain are treated as different from hostname, so a separate entry is required for each.
+
+
+## Troubleshooting Authentication
+
+
+###The Process:
+
+- the omi client requests authentication using the username and optional password to acquire a stored credential.
```
jeffaco: Capitalize "t" of "The", it's effectively a sentence.

Unix/installbuilder/conf/omiserver.conf
```diff
@@ -46,3 +46,14 @@ keyfile=/etc/opt/omi/ssl/omikey.pem
...
+# of the credentials file. This file is in the form
+#
+# <domain>|<username>:<password> 
+#
+# For further information, take a look at the file Unix/doc/setup-ntlm-omi.md
+#
+#
+#NtlmCredsFile=PATH
+# For example
+NtlmCredsFile=/etc/opt/omi/.creds/ntlm
```
jeffaco: Please move this segment to a slightly more obvious block, like just after 22, putting a blank line after this segment (to separate from other segments).

Unix/installbuilder/conf/omiserver.conf
```diff
@@ -46,3 +46,14 @@ keyfile=/etc/opt/omi/ssl/omikey.pem
 configfile=/etc/opt/omi/conf/omiserver.conf
 #NoSSLv2=true
 #NoSSLv3=false
+# To enable NTLM authentication via negotiation (SPNEGO), specify the location 
+# of the credentials file. This file is in the form
+#
+# <domain>|<username>:<password> 
+#
+# For further information, take a look at the file Unix/doc/setup-ntlm-omi.md
```
jeffaco: You're assuming that the reader is familiar with GitHub, or with the source tree. Most OMI customers won't be familiar with either. I recommend something like:

`# For further information, please see link: https://github.com/Microsoft/omi/blob/master/Unix/doc/setup-ntlm-omi.md`

Issue #221, opened 2017-02-06
Chunking tests still fail on Mac


palladia:

palladia:
Fixed by #223

PR #223, opened 2017-02-06
Issue #221:  Also check for errno 35

palladia:


palladia:
@Microsoft/ostc-devs

Unix/tests/http/test_http.cpp
```diff
@@ -730,12 +732,14 @@ static void _ConnectToServerExpectConnectionDrop(const string& data, MI_Uint32 s
...
         size_t read = 0;
+        int err;
 
         do
         {
             r = Sock_Read(sock, r_buf, sizeof(r_buf), &read);
+            err = Sock_GetLastError();
         }
-        while (r != MI_RESULT_OK && Sock_GetLastError() == 11 /*EAGAIN*/);
+        while (r != MI_RESULT_OK && (err == 11 || err == 35) /*EAGAIN*/);
```
jeffaco: Err 11 is EGAGIN, not 35. Can you clarify what 35 is, here and below?

Unix/tests/http/test_httpclient.cpp
```diff
@@ -395,12 +395,14 @@ static void* MI_CALL _http_server_proc(void* param)
...
     size_t read = 0;
+    int err;
 
     do
     {
         r = Sock_Read(sock, r_buf, sizeof(r_buf), &read);
+        err = Sock_GetLastError();
     }
-    while (r != MI_RESULT_OK && Sock_GetLastError() == 11 /*EAGAIN*/);
+    while (r != MI_RESULT_OK && (err == 11 || err == 35) /*EAGAIN*/);
```
jeffaco: Ditto to above, this is wrong from a comment perspective.

Unix/tests/wsman/test_wsmanredirect.cpp
```diff
@@ -107,12 +107,15 @@ static void* MI_CALL _http_server_proc(void* param)
...
-    
+    int err;
+
     do
     {
         r = Sock_Read(sock, r_buf, sizeof(r_buf), &read);
+        err = Sock_GetLastError();
     }
-    while (r != MI_RESULT_OK && Sock_GetLastError() == 11 /*EAGAIN*/);
+    while (r != MI_RESULT_OK && (err == 11 || err == 35) /*EAGAIN*/);
```
jeffaco: and ditto ...

Unix/tests/http/test_http.cpp
paulcallen: I believe if you include errno.h you can use the actual defines for the different errors which is more explanitory

jeffaco: That's actually much better (use the #define value). Allows for self-documenting code ...

Unix/tests/http/test_http.cpp
```diff
@@ -185,14 +185,16 @@ static void* MI_CALL _http_client_proc(void* param)
...
     size_t read = 0;
+    int err;
 
     do
     {
         r = Sock_Read(sock, r_buf, sizeof(r_buf), &read);
+        err = Sock_GetLastError();
     }
-    while (r != MI_RESULT_OK && Sock_GetLastError() == 11 /*EAGAIN*/);
+    while (r != MI_RESULT_OK && (err == 11 || err == 35) /*EAGAIN and EDEADLK*/);
```
jeffaco: This isn't quite what I had in mind.

Can't you include errlog.h and then do something like:

`while (r != MI_RESULT_OK && (err == EAGAIN || err == EDEADLK));`

That's more self documenting ...

palladia: I don't think error codes (from errno.h) is standardized for all platforms.  What's confusing is that errno of 35 could be EAGAIN, ENOMSG, EDEADLK, or undefined, just on a few systems that I checked.

jeffaco: That doesn't make sense, especially for the lower numbered (standard) error codes.

Can you please point out two systems that conflict with one another so I can take a look, thanks.

palladia:
@jeffaco Please re-review.

PR #224, opened 2017-02-07
Fine edits to doc and omiserver.conf before final release

yakman2020:
@Microsoft/omi-devs

### DIFF
--- a/Unix/installbuilder/conf/omiserver.conf
+++ b/Unix/installbuilder/conf/omiserver.conf
@@ -20,6 +20,18 @@ httpsport=0
 ##
 #trace=(true|false)
 
+## To enable NTLM authentication via negotiation (SPNEGO), specify the location 
+## of the credentials file. This file is in the form
+##
+## <domain>|<username>:<password> 
+##
+## For further information, please see link: https://github.com/Microsoft/omi/blob/master/Unix/doc/setup-ntlm-omi.md
+##
+##
+##NtlmCredsFile=PATH
+## For example
+#NtlmCredsFile=/etc/opt/omi/.creds/ntlm
+
 ##
 ## <NICKNAME> -- set the value of nickname.
 ##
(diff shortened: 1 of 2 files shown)


Unix/installbuilder/conf/omiserver.conf
```diff
@@ -20,6 +20,18 @@ httpsport=0
...
+## of the credentials file. This file is in the form
+##
+## <domain>|<username>:<password> 
+##
+## For further information, please see link: https://github.com/Microsoft/omi/blob/master/Unix/doc/setup-ntlm-omi.md
+##
+##
+##NtlmCredsFile=PATH
+## For example
+NtlmCredsFile=/etc/opt/omi/.creds/ntlm
```
paulcallen: you are checking in the actual cred file location? I think this is just an example that is not commented out

jeffaco: Oh, good catch. For what's checked in, this should be commented out (right Paul?).

The customer should enable this only if they wanted to.

I'd also get rid of the example, change to something like:

```
## To enable NTLM authentication via negotiation (SPNEGO), specify the location 
## of the credentials file. This file is in the form
##
## <domain>|<username>:<password> 
##
## For further information, please see link: https://github.com/Microsoft/omi/blob/master/Unix/doc/setup-ntlm-omi.md
##
#NtlmCredsFile=/etc/opt/omi/.creds/ntlm
```

yakman2020:
done

Unix/installbuilder/conf/omiserver.conf
```diff
@@ -20,6 +20,18 @@ httpsport=0
...
+## of the credentials file. This file is in the form
+##
+## <domain>|<username>:<password> 
+##
+## For further information, please see link: https://github.com/Microsoft/omi/blob/master/Unix/doc/setup-ntlm-omi.md
+##
+##
+##NtlmCredsFile=PATH
+## For example
+#NtlmCredsFile=/etc/opt/omi/.creds/ntlm
```
jeffaco: I think you missed my prior comment. Please delete lines 30-32, you don't need them.

PR #232, opened 2017-02-10
Modify configure script to abort if SSL 1.0.x not installed on AIX

jeffaco:
Verify that SSL 1.0.x is installed on AIX platforms.
Also cleaned up configure output (commented on code review, but corrections not made).

### DIFF
--- a/Unix/configure
+++ b/Unix/configure
@@ -1344,6 +1344,46 @@ rm -f $tmpdir/pthread_rwlock_t_func
 
 ##==============================================================================
 ##
+## Check whether SSL 1.0.x is installed on AIX platforms
+##
+##==============================================================================
+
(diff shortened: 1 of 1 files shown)


Unix/configure
```diff
@@ -1344,11 +1344,50 @@ rm -f $tmpdir/pthread_rwlock_t_func
 
 ##==============================================================================
 ##
+## Check whether SSL 1.0.x is installed on AIX platforms
+##
+##==============================================================================
+
+echo $echon "checking if SSL 1.0.x is installed on AIX platform ... $echoc"
```
paulcallen: shouldn't the echo be inside the if statement so it is not echoed on other platforms?

jeffaco: No. I always echo that on all platforms, but non AIX platforms say "N/A". It was most consistent with everything else in configure (we never had a platform-specific test like this).

PR #239, opened 2017-02-24
Migrate httpsport and sslciphersuite values from old omiserver.conf

johnkord:


johnkord:
@Microsoft/ostc-devs

johnkord:
Upgrading on debian systems will prompt the user with: 

Configuration file '/etc/opt/omi/conf/omiserver.conf'
 ==> Modified (by you or by a script) since installation.
 ==> Package distributor has shipped an updated version.
   What would you like to do about it ?  Your options are:
    Y or I  : install the package maintainer's version
    N or O  : keep your currently-installed version
      D     : show the differences between the versions
      Z     : start a shell to examine the situation
 The default action is to keep your current version.
*** omiserver.conf (Y/I/N/O/D/Z) [default=N] ?


The default (keep current version) works just fine.  In the future, we might want to manage our own conf files instead of having them managed by the native packagers.  I think it adds more complexity than it is worth

PR #240, opened 2017-03-01
Make SSL 0.9.8 optional in ulinux build

paulcallen:
@Microsoft/omi-devs

### DIFF
--- a/Unix/configure
+++ b/Unix/configure
@@ -316,6 +316,11 @@ do
         disable_makefile_gen=1
         ;;
 
+    --disable-ssl-0.9.8)
+        disable_ssl_0_9_8=1
+        ;;
+
+
     *)
         echo "$0: unknown option:  $opt"
         exit 1
@@ -413,7 +418,8 @@ OPTIONS:
                             builds of OMI, including ULINUX as necessary
     --disable-makefile-gen  Disable Makefile generation (used internally for
                             universal builds)
-
+    --disable-ssl-0.9.8     When building with --enable-ulinux, skip the ssl
+                            0.9.8 build and only do ssl 1.0.0.
 EOF
     exit 0
 fi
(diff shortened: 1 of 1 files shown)


paulcallen:
@Microsoft/omi-devs Can I get a code review for this please. It is needed for the pending release.

Unix/configure
```diff
@@ -413,7 +418,8 @@ OPTIONS:
                             builds of OMI, including ULINUX as necessary
     --disable-makefile-gen  Disable Makefile generation (used internally for
                             universal builds)
-
+    --skip-ssl-0.9.8        When building with --enable-ulinux, skip the ssl
```
jeffaco: I would prefer, for consistency, --disable-ssl-0.,9.8

Unix/configure
```diff
@@ -2468,13 +2476,20 @@ OMI_CONFIGURE_QUALS=--enable-microsoft --disable-makefile-gen
 
 all:
 ifeq (\$(ENABLE_ULINUX),1)
+EOF
```
jeffaco: I prefer you not do the EOF thing to write a fragment of the GNUmakefile file. This will make it harder to change GNUmakefile down the road. Just write the variables out (like they currently are, for ENABLE_ULINUX), and then test against them.

Unix/configure
```diff
@@ -494,12 +500,14 @@ if [ "$enable_ulinux" = "1" ]; then
     openssl100_cflags=`PKG_CONFIG_PATH=$openssl100_libdir/pkgconfig $pkgconfig --cflags openssl`
     openssl100_libs=`PKG_CONFIG_PATH=$openssl100_libdir/pkgconfig $pkgconfig --libs openssl`
 
+if [ "$disable_ssl_0_9_8" != "1" ]; then
```
jeffaco: This should be indented properly to match what's above. Like C++, it's not required, but you should have it for readability.

PR #243, opened 2017-03-06
Update README.md to link to ntlm setup page

yakman2020:
This is an immediate help for documentation issues.  More to follow, but this should help


yakman2020:
Done.

PR #256, opened 2017-03-21
Implement create/modify/delete operation in Color Provider

JumpingYang001:
I have implemented create/modify/delete operation in Color Provider, that I want to test these 3 operations and will add more in the future, can you help to review this PR? thanks very much!


JumpingYang001:
Ho, we have exist provider for testing, just close it, thanks.

https://github.com/Microsoft/omi/blob/master/Unix/samples/Providers/President/PresidentLink.c 
Associator  Reference Enumerate 

https://github.com/Microsoft/omi/blob/master/Unix/samples/Providers/TestClass_Employee/Employee.cpp
Enumerate GetInstance CreateInstance ModifyInstance DeleteInstance

Issue #257, opened 2017-03-22
release builds should not include debug information on any platform


paulcallen:
* Solaris definitely has debug symbols
* We removed symbols previously on Linux, make sure we did it correctly though
* Make sure other Unix platforms don't have symbols

paulcallen:
OSX also builds release with debug symbols.

PR #261, opened 2017-03-24
Remove debug symbols for release builds

paulcallen:
@Microsoft/omi-devs

### DIFF
--- a/Unix/buildtool
+++ b/Unix/buildtool
@@ -811,8 +811,8 @@ if [ "$arg1" = "cflags" -o "$arg1" = "cxxflags" ]; then
 
             ## Optimization options.
             if test -z "$debug_opt"; then
-                test -n "$c_opt" && r="$r -g"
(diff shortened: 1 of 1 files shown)


paulcallen:
fixes #257 
Sun and OSX both had symbols in release build.

Unix/buildtool
```diff
@@ -811,8 +811,8 @@ if [ "$arg1" = "cflags" -o "$arg1" = "cxxflags" ]; then
 
             ## Optimization options.
             if test -z "$debug_opt"; then
-                test -n "$c_opt" && r="$r -g"
-                test -n "$cxx_opt" && r="$r -g -O"
+                test -n "$c_opt" && r="$r"
```
jeffaco: Huh? You're saying, for the C compiler, make `r="$r"`. This won't do anything at all, will it?

paulcallen: was just doing the safe option of removing the -g :)

Issue #231, opened 2017-02-10
NtlmCredsFile form should be <domain>:<username>:<password>  in omiserver.conf


JumpingYang001:
File:https://github.com/Microsoft/omi/blob/dcc0a298add923d587630226957b43cd8b5f97fc/Unix/installbuilder/conf/omiserver.conf
Current:`<domain>|<username>:<password> `
Expect:`<domain>:<username>:<password>`

JumpingYang001:
Fixed the issue on #263 .

PR #263, opened 2017-03-27
Fix issue 231

JumpingYang001:
I have fixed a small typo issue #231 , could you help to review this PR? thanks very much!


jeffaco:
I merged this myself after fixing your commit message.

Commit messages should be <= 70 bytes in length, and should not include questions, just state what was done (and why if not obvious).

If this comes up often, I'll give you direct access to the repo. However, please follow commit message guidelines. For now, though, we can just see how often this comes up.

Issue #262, opened 2017-03-24
fix configure/buildtool scripts to properly support --enable-gcov


paulcallen:
buildtool needs to add the following compiler options if enable-gcov was specified:
-fprofile-arcs -ftest-coverage 
We need to do this by passing a new option through to buildtool if enable_gcov=1, and then if that was passed in add these options

according to Jumping, link flags also need to be added too, he added link options:
-fprofile-arcs -ftest-coverage  -lgcov

PR #266, opened 2017-03-29
Enable gcov to support --enable-gcov 

JumpingYang001:
I have implemented issue #262 to support --enable-gcov , could you help to review this PR? thanks very much!

### DIFF
--- a/Unix/buildtool
+++ b/Unix/buildtool
@@ -738,6 +738,9 @@ if [ "$arg1" = "cflags" -o "$arg1" = "cxxflags" ]; then
             --debug)
                 debug_opt=1
                 ;;
+            --gcov)
+                gcov_opt=1
+                ;;
             --pic)
                 pic_opt=1
                 ;;
@@ -776,6 +779,7 @@ if [ "$arg1" = "cflags" -o "$arg1" = "cxxflags" ]; then
                 fi
             fi
 
+            test -n "$gcov_opt" && r="$r -ftest-coverage -fprofile-arcs"
             test -n "$pic_opt" && r="$r -fPIC"
             test -n "$errwarn_opt" && r="$r -Werror"
             r="$r -Wall"
--- a/Unix/mak/rules.mak
+++ b/Unix/mak/rules.mak
@@ -121,6 +121,10 @@ ifeq ($(ENABLE_WERROR),1)
   __OPTS += --errwarn
 endif
 
+ifeq ($(ENABLE_GCOV),1)
+  __OPTS += --gcov
+endif
+
 ifeq ($(CONFIG_FAVORSIZE),1)
   __OPTS += --size
 endif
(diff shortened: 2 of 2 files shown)


Unix/mak/rules.mak
```diff
@@ -176,6 +180,15 @@ CXXPROGFLAGS+=$(shell $(BUILDTOOL) syslibs)
...
 CXXPROGFLAGS+=$(LIBPATHFLAGS)
 
+ifeq ($(ENABLE_GCOV),1)
+  CFLAGS+=-lgcov
+  CPROGFLAGS+=-lgcov
+  CSHLIBFLAGS+=-lgcov
+  CXXFLAGS+=-lgcov
+  CXXPROGFLAGS+=-lgcov
+  CXXSHLIBFLAGS+=-lgcov
+endif
```
jeffaco: I don't think this is right, and would like a new iteration.

The `-l` qualifier is for linking in a library. So, if you say `-lgcov`, you're saying include library libgcov.{so|a} with this.

The `-l` qualifier is a LINKER flag, not a COMPILER flag. So it doesn't belong in CFLAGS, which is specifically for compiling. It may work (if CFLAGS is included when linking), but it's not where this belongs.

Please resolve this by putting it in the right place (should just be one or two variables), then posting a new iteration.

Unix/buildtool
```diff
@@ -776,6 +779,8 @@ if [ "$arg1" = "cflags" -o "$arg1" = "cxxflags" ]; then
                 fi
             fi
 
+			test -n "$gcov_opt" && r="$r -ftest-coverage -fprofile-arcs -lgcov"
```
paulcallen: Not sure you need the -lgcov here as that us a link option and not compile options.

JumpingYang001: Fixed.

Unix/mak/rules.mak
JumpingYang001: Fixed, and addition information why I add -lgcov in the other 4 flags:
```
If I don't add -lgcov in CSHLIBFLAGS and CXXSHLIBFLAGS, it will fail with below error:

/usr/bin/ld: /test/omi/Unix/output/bin/nits: hidden symbol `__gcov_init' in /usr/lib/gcc/x86_64-redhat-linux/4.8.5/libgcov.a(_gcov.o) is referenced by DSO
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status


If I don't add -lgcov in CPROGFLAGS and CXXPROGFLAGS, it will fail with below error:

g++ -o /test/omi/Unix/output/bin/chkshlib /test/omi/Unix/output/obj/chkshlib/chkshlib.o -L/test/omi/Unix/output/lib -L/usr/lib64      -lpthread -ldl -lpam -lssl -lcrypto    -Wl,-rpath=/opt/omi-1.0.8/lib -Wl,-rpath=/usr/lib64
/test/omi/Unix/output/obj/chkshlib/chkshlib.o: In function `_GLOBAL__sub_I_65535_0_chkshlib.cpp':
chkshlib.cpp:(.text.startup+0xd0): undefined reference to `__gcov_init'
/test/omi/Unix/output/obj/chkshlib/chkshlib.o:(.data.rel+0x20): undefined reference to `__gcov_merge_add'
collect2: error: ld returned 1 exit status
```

jeffaco:
Merged to master branch.

PR #270, opened 2017-04-03
Jeff packaging

jeffaco:

Please review for packaging changes; this should name packages (for Linux) consistently with omi-kits repository.

This will also not build a kit by default unless you explicitly ask for it during configure.

### DIFF
--- a/Unix/configure
+++ b/Unix/configure
@@ -75,7 +75,6 @@ set_microsoft_build_options()
     sysconfdir=/etc/opt/omi/conf
     certsdir=/etc/opt/omi/ssl
     credsdir=/etc/opt/omi/creds
-    enable_native_kits=1
 }
 
 ##==============================================================================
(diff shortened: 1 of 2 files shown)


Unix/configure
```diff
@@ -2471,9 +2470,14 @@ __ROOT=$root
 export OUTPUTDIR=$outputdir
 export ROOT=$root
 export ENABLE_ULINUX=$enable_ulinux
+export ENABLE_NATIVE_KITS=$enable_native_kits
 export DISABLE_SSL_0_9_8=$disable_ssl_0_9_8
 
 OMI_CONFIGURE_QUALS=--enable-microsoft --disable-makefile-gen
+ifeq (\$(ENABLE_NATIVE_KITS),1)
+  OMI_CONFIGURE_QUALS=--enable-microsoft --disable-makefile-gen --enable-native-kits
```
paulcallen: the above if statement says 'if we are not disabling makefile gen', then we add 'disable makefile gen' to the options. Or did I misunderstand something?

jeffaco: So this is a generated Makefile (generated by configure). What this says is: If, on the initial configure line, you did a `--enable-native-kits`, pass that on down if you need to reconfigure for SSL 0.9.8 and SSL 1.0.0.

PR #272, opened 2017-04-03
Compress kits on HP-UX and Solaris platforms

jeffaco:
Most platforms compress native kits. However, HP-UX and Solaris do not.
Compress kits on those platforms so they're a reasonable size.


jeffaco:
Spacing is correct, and renders properly using Chrome. I think that was an Edge-ism.

PR #276, opened 2017-04-04
Add build dependencies to shared libraries

### LINKED ISSUE #275: Add proper dependency checking for shared libraries in the makefiles
we only have dependencies for programs today.

paulcallen:
Currently there are only dependencies for programs.
Fixes #275.

@Microsoft/omi-devs

### DIFF
--- a/Unix/mak/rules.mak
+++ b/Unix/mak/rules.mak
@@ -180,6 +180,8 @@ CXXPROGFLAGS+=$(shell $(BUILDTOOL) syslibs)
 CXXPROGFLAGS+=$(OPENSSL_LIBS)
 CXXPROGFLAGS+=$(LIBPATHFLAGS)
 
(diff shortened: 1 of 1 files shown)


Unix/mak/rules.mak
```diff
@@ -275,12 +275,13 @@ TARGET = $(LIBDIR)/$(shell $(BUILDTOOL) shlibname $(CSHLIBRARY))
 
 __LIBRARIES = $(addprefix -l,$(LIBRARIES))
 __LIBRARIES += $(addprefix -l,$(SYSLIBRARIES))
+__DEPS=$(wildcard $(addprefix $(LIBDIR)/lib,$(addsuffix .*,$(LIBRARIES))))
```
jeffaco: This is consistent for both `CSHLIBRARY` and `CXXSHLIBRARY`. Rather than copy/paste, you might want to consider just defining this rule once.

PR #277, opened 2017-04-04
Deprecate wchar builds for regress and configure

paulcallen:
Disable wchar option on configure and regress so we no longer validate
this build flavor. This is an unused flavor of OMI so there is no point
in maintaining compatibility. It has been left in the files as commented
out in case we do need to re-enable it.
Fixes issue #273 
@Microsoft/omi-devs

### DIFF
--- a/Unix/configure
+++ b/Unix/configure
@@ -166,10 +166,6 @@ do
       enable_werror=1
       ;;
 
-    --enable-wchar)
-      enable_wchar=1
-      ;;
-
     --enable-gcov)
       enable_gcov=1
       ;;
@@ -354,7 +350,6 @@ OPTIONS:
     --disable-rtti          Disable C++ RTTI support.
     --disable-templates     Disable C++ templates support.
     --enable-werror         Treat warnings as errors.
-    --enable-wchar          Use 'wchar_t' character type [char].
     --target=TARGET         Cross-compiler for the given platform.
     --prefix=PATH           The installation prefix [/opt/omi-$version]
     --libdir=PATH           Install library components here [*/lib].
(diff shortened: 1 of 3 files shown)


Unix/configure
```diff
@@ -166,9 +166,10 @@ do
       enable_werror=1
       ;;
 
-    --enable-wchar)
-      enable_wchar=1
-      ;;
+## Disabling support for wchar builds
+##    --enable-wchar)
+##      enable_wchar=1
+##      ;;
```
jeffaco: In general, I'm not a fan of this. Git is a revision control system, and it's trivial to revert a commit. Rather than leaving out commented sections here, removing 257 below, etc, I prefer just getting rid of it with an obvious commit message. Then we can find it and revert it if it was a mistake.

Leaving this here just makes people wonder what's going on. And you'd need to revert to get 257 back anyway. So rather than have historical commented out sections, keep it clean and revert later if needed.

Unix/regress
```diff
@@ -182,25 +181,26 @@ fi
...
+##
+##if [ "$?" != "0" ]; then
+##    echo "$0: failed 'wchar' tests"
+##    cleanup  $current_output
+##    exit 1
+##fi
+##
+##sudo rm -rf ./$current_output $new_output GNUmakefile
+##
+##fi
```
jeffaco: Strictly speaking, if `$run_wchar` is always zero, you don't need to comment this out at all. But I still recommend removing it to keep it clean, shorter, and thus easier to understand.

Unix/configure
paulcallen: I guess I didn't want to purge everything to do with wchar from both the files as it was a bigger change. I will re-evaluate.

paulcallen:
#jeffaco I have updated to remove all wchar stuff from configure and regress. Thoughts?

PR #279, opened 2017-04-05
Fix #254 about inner message in ProviderFault

### LINKED ISSUE #254: another windows error that is not decoded well
did this:
`./omicli --hostname <windows-box>  -u <user> -p <password>  --auth Basic  --encryption none gi root/cimv2 { Win32_Process Handle 1 }`
`./omicli: result: MI_RESULT_FAILED`
`./omicli: result: Internal error parsing Wsman fault message`

Process ID 1 on this windows box does not exist. Should have resulted in MI_RESULT_NOT_FOUND with the appropriate extracted string from the packet.

<s:Envelope xml:lang="en-US" xmlns:s="http://www.w3.org/2003/05/soap-envelope" xmlns:a="http://schemas.xmlsoap.org/ws/2004/08/addressing" xmlns:x="http://schemas.xmlsoap.org/ws/2004/09/transfer" xmlns:e="http://schemas.xmlsoap.org/ws/2004/08/eventing" xmlns:n="http://schemas.xmlsoap.org/ws/2004/09/enumeration" xmlns:w="http://schemas.dmtf.org/wbem/wsman/1/wsman.xsd" xmlns:p="http://schemas.microsoft.com/wbem/wsman/1/wsman.xsd"><s:Header><a:Action>http://schemas.xmlsoap.org/ws/2004/08/addressing/fault</a:Action><a:MessageID>uuid:3679CAE4-326C-485F-9BC9-033FE51D4CEE</a:MessageID><a:To>http://schemas.xmlsoap.org/ws/2004/08/addressing/role/anonymous</a:To><a:RelatesTo>uuid:6E063D19-4AF3-0005-0000-000000010000</a:RelatesTo></s:Header><s:Body><s:Fault><s:Code><s:Value>s:Sender</s:Value><s:Subcode><s:Value>a:DestinationUnreachable</s:Value></s:Subcode></s:Code><s:Reason><s:Text xml:lang="en-US">The WS-Management service cannot process the request. The service cannot find the resource identified by the resource URI and selectors. </s:Text></s:Reason><s:Detail><w:FaultDetail>http://schemas.dmtf.org/wbem/wsman/1/wsman/faultDetail/InvalidResourceURI</w:FaultDetail><f:WSManFault xmlns:f="http://schemas.microsoft.com/wbem/wsman/1/wsmanfault" Code="2150858752" Machine="1.2.3.4"><f:Message><f:ProviderFault provider="WMI Provider" path="C:\Windows\system32\WsmWmiPl.dll"><f:WSManFault xmlns:f="http://schemas.microsoft.com/wbem/wsman/1/wsmanfault" Code="2150858752" Machine="paulall4.redmond.corp.microsoft.com"><f:Message>The WS-Management service cannot process the request. The service cannot find the resource identified by the resource URI and selectors. </f:Message></f:WSManFault><f:ExtendedError><p:__ExtendedStatus xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:p="http://schemas.dmtf.org/wbem/wscim/1/cim-schema/2/__ExtendedStatus" xmlns:cim="http://schemas.dmtf.org/wbem/wscim/1/common" xsi:type="p:__ExtendedStatus_Type"><p:Description xsi:nil="true" xsi:type="cim:cimString"/><p:Operation xsi:type="cim:cimString">GetObject</p:Operation><p:ParameterInfo xsi:type="cim:cimString">Win32_Process.Handle=&quot;2&quot;</p:ParameterInfo><p:ProviderName xsi:type="cim:cimString">CIMWin32</p:ProviderName><p:StatusCode xsi:nil="true" xsi:type="cim:cimUnsignedInt"/></p:__ExtendedStatus></f:ExtendedError></f:ProviderFault></f:Message></f:WSManFault></s:Detail></s:Fault></s:Body></s:Envelope>

JumpingYang001:
@Microsoft/omi-devs , could you help to review this PR? thank very much!

### DIFF
--- a/Unix/tests/wsman/test_wsmanparser.cpp
+++ b/Unix/tests/wsman/test_wsmanparser.cpp
@@ -1377,13 +1377,14 @@ NitsTestWithSetup(TestPullResponse2, TestParserSetup)
 }
 
 NitsEndTest
-NitsTestWithSetup(TestFaultResponse2, TestParserSetup)
+NitsTestWithSetup(TestInternalErrorFaultResponse, TestParserSetup)
 {
     WSMAN_WSHeader wsheaders;
     XML xml;
     WSMAN_WSFault fault;
     Error_Types errorType;
     MI_Char *errorTypeStr;
+    MI_Result isGetWsmanError;
 
     XML_Char data[] = PAL_T("<s:Envelope xml:lang=\"en-US\" xmlns:s=\"http://www.w3.org/2003/05/soap-envelope\" ")
         PAL_T("xmlns:a=\"http://schemas.xmlsoap.org/ws/2004/08/addressing\" ")
@@ -1437,6 +1438,7 @@ NitsTestWithSetup(TestFaultResponse2, TestParserSetup)
     {
         goto cleanup;
     }
+    
     if (!NitsCompare(wsheaders.rqtAction, WSMANTAG_ACTION_FAULT_WSMAN, PAL_T("Action is not Fault Response")))
     {
         goto cleanup;
@@ -1446,37 +1448,45 @@ NitsTestWithSetup(TestFaultResponse2, TestParserSetup)
     {
         goto cleanup;
     }
+    
     if (!NitsCompareString(fault.code, PAL_T("http://www.w3.org/2003/05/soap-envelope:Receiver"), 
                            PAL_T("Mismatch of fault code")))
     {
         goto cleanup;
     }
+    
     if (!NitsCompareString(fault.subcode, 
                            PAL_T("http://schemas.dmtf.org/wbem/wsman/1/wsman.xsd:InternalError"), 
                            PAL_T("Mismatch of fault subcode")))
     {
         goto cleanup;
     }
+    
     if (!NitsCompareString(fault.reason, NULL, PAL_T("Mismatch of fault reason")))
     {
         goto cleanup;
     }
+    
     if (!NitsCompareString(fault.detail, PAL_T("Error with error code -2144108472 occurred while calling method WSManPluginReceiveResult."), 
                            PAL_T("Invalid fault detail")))
     {
         goto cleanup;
     }
-    if (!NitsCompare(MI_RESULT_OK, GetWsmanErrorFromSoapFault(fault.code, fault.subcode, fault.detail, &errorType, &errorTypeStr),
+    
+    isGetWsmanError = GetWsmanErrorFromSoapFault(fault.code, fault.subcode, fault.detail, &errorType, &errorTypeStr);
+    
+    if (!NitsCompare(MI_RESULT_OK, isGetWsmanError,
                      PAL_T("Cannot determine errorCode")))
     {
         goto cleanup;
     }
+    
     if (!NitsCompare(ERROR_INTERNAL_ERROR, errorType, PAL_T("Wrong error type")))
     {
         goto cleanup;
     }
 
-    if (!NitsCompare(0, fault.mi_result, PAL_T("Wrong result code")))     // MI_RESULT_NO_SUCH_PROPERTY
+    if (!NitsCompare(MI_RESULT_FAILED, fault.mi_result, PAL_T("Wrong result code")))
     {
         goto cleanup;
     }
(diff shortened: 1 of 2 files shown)


Unix/tests/wsman/test_wsmanparser.cpp
```diff
@@ -1487,6 +1487,138 @@ NitsTestWithSetup(TestFaultResponse2, TestParserSetup)
...
+                           PAL_T("http://schemas.xmlsoap.org/ws/2004/08/addressing:DestinationUnreachable"), 
+                           PAL_T("Mismatch of fault subcode")))
+    {
+        goto cleanup;
+    }
+    if (!NitsCompareString(fault.reason, PAL_T("The WS-Management service cannot process the request. The service cannot find the resource identified by the resource URI and selectors. "), PAL_T("Mismatch of fault reason")))
+    {
+        goto cleanup;
+    }
+    if (!NitsCompareString(fault.detail, PAL_T("CIMWin32"),
```
paulcallen: This does not seem right, especially as there is nothing useful in the details. Understandably the extended status->description is empty, which is not helpful, but maybe this string should default to the fault.reason if that is the case.

Unix/tests/wsman/test_wsmanparser.cpp
```diff
@@ -1487,6 +1487,138 @@ NitsTestWithSetup(TestFaultResponse2, TestParserSetup)
...
+                     PAL_T("Cannot determine errorCode")))
+    {
+        goto cleanup;
+    }
+    if (!NitsCompare(ERROR_WSMAN_DESTINATION_UNREACHABLE, errorType, PAL_T("Wrong error type")))
+    {
+        goto cleanup;
+    }
+
+    if (!NitsCompare(0, fault.mi_result, PAL_T("Wrong result code")))
```
paulcallen: This does not seem a correct mapping either, The operation failed, and yet we got an MI_RESULT_OK from it. This should either map to MI_RESULT_FAILED, or a more appropriate error if one exists.

paulcallen: In fact I see this is the same as the previous unittest, and that does not seem correct either. I think there may have been a bug in the code from the previous change that needs fixing

JumpingYang001: Fixed. In fact, it is correct and the real error is stored in errorTypeStr, but it is confused, so I change it to be more understandable.

Unix/tests/wsman/test_wsmanparser.cpp
JumpingYang001: Fixed.

Unix/tests/wsman/test_wsmanparser.cpp
JumpingYang001: Ho, do you mean fault.mi_result should not be MI_RESULT_OK?

JumpingYang001: We didn't set correct fault->mi_result in WS_ParseFaultBody, I will fix it.

JumpingYang001: Fixed.

JumpingYang001:
I have fixed the issues, thanks.

Unix/wsman/wsmanparser.c
```diff
@@ -3241,29 +3240,66 @@ int WS_ParseFaultBody(
                 }
                 else if (Tcscmp(e.data.data, ZT("WSManFault")) == 0)
                 {
+                    /* fault->mi_result should be MI_RESULT_FAILED for WSManFault */
+                    fault->mi_result = MI_RESULT_FAILED;
```
paulcallen: Ah, so here is the catch here! The fault result code needs to get matched to the most appropriate error code for the fault. There is some code somewhere for this. I will take a look in a while to see if I can track it down. Otherwise a TIMEOUT error will not result in a MI_RESULT_TIMEOUT error.

paulcallen: OK, so I did not find a mapping when I thought there was. I thought it was done in wsmanerrorhandling.c where it does most of the error processing.
For now lets leave it the way you did it.

JumpingYang001:
Thanks for reviewing!

PR #284, opened 2017-04-12
force compress, without -f it blocks compress operation when .Z file …

BingyuLiu:
…has been there.


johnkord:
Spacing is a little off, if you can fix that before checkin, but otherwise, I sign off!

PR #285, opened 2017-04-13
Fixes issue with decrypt on ppc rhel.

yakman2020:


yakman2020:
@Microsoft/omi-devs

yakman2020:
pbuild completed.

Unix/http/httpauth.c
```diff
@@ -564,23 +564,25 @@ MI_Boolean Http_DecryptData(_In_ Http_SR_SocketData * handler, _Out_ HttpHeaders
...
+    memcpy(buffer_p, original_encoding, strlen(original_encoding) + 1);
+    pHeaders->charset = buffer_p;
 
-    pHeaders->contentType = original_content_type;
     pHeaders->contentLength = original_content_length;
-    pHeaders->charset = original_encoding;
+
+    page->u.s.size = output_buffer.length;
+    memcpy(page + 1, output_buffer.value, output_buffer.length);
+    (*_g_gssState.Gss_Release_Buffer)(&min_stat, &output_buffer);
```
paulcallen: ok, so something does not look right here. buffer_p is a pointer into output_buffer. pHeaders->* get pointers into that buffer. Then you release the buffer and return and I am assuming callers are going to use what is in pHeaders.
Therefore the caller is going to get garbled use-after-free pointers.
Of course I may be misreading this and your code is correct....

PR #287, opened 2017-04-13
Fix SLIST on PPC due to memory cache problems

paulcallen:
Memory ordering on PPC is different to X86/AMD and so just because one
chunk of memory is synchronized does not mean others are. This change
adds a memory block around the locking and unlocking of the SLIST so all
relevant memory in the SLIST is properly pulled into the processor
memory cache.

@Microsoft/omi-devs


paulcallen:
Fixes #288

PR #289, opened 2017-04-14
Fix AIX libgssapi library loading

yakman2020:
We are currently failing to load the libgssapi_krb5 library because AIX uses shared objects in an archive file for some reason. This requires different flags and file name to load.  With this fix, AIX loads the file correctly.


yakman2020:
@Microsoft/omi-devs

PR #291, opened 2017-04-18
Add no-ssl-compression option to configuration

paulcallen:
@Microsoft/omi-devs

### DIFF
--- a/Unix/http/httpcommon.c
+++ b/Unix/http/httpcommon.c
@@ -214,6 +214,14 @@ MI_Result CreateSSLContext(SSL_CTX **sslContext, SSL_Options sslOptions)
         options |= SSL_OP_NO_TLSv1_2;
     }
 #endif
+    if ( sslOptions & DISABLE_SSL_COMPRESSION)
+    {
+#ifdef SSL_OP_NO_COMPRESSION
+        options |= SSL_OP_NO_COMPRESSION;
+#else
+        trace_Http_SslCompressionNotPresent();
+#endif
+    }
     if ( options != 0)
     { 
         // If options is zero, the operation is a noop. SSL_CTX_set_options only sets, never clears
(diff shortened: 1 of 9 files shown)


Unix/http/httpcommon.c
```diff
@@ -214,6 +214,12 @@ MI_Result CreateSSLContext(SSL_CTX **sslContext, SSL_Options sslOptions)
         options |= SSL_OP_NO_TLSv1_2;
     }
 #endif
+#ifdef SSL_OP_NO_COMPRESSION
```
jeffaco: What sets `SSL_OP_NO_COMPRESSION`? I don't see anything in this checkin that does that.

Here, you do an `#ifdef`, but you don't do that in other places (cli_c.c, options.c, server.c). I'd like to see consistency; either have an `#ifdef`, and be consistent, or don't have an `#ifdef` anywhere.

In general, if there's never a reason to disable an option, I'm not a fan of `#ifdef`, it just increases complexity and testing matrix. Is there a reason to disable this? If not, please get rid of it. If so, then add `#ifdef` tests to other usages.

paulcallen: SSL_OP_NO_COMPRESSION  is an option from the SSH header file itself, in much the same way that SSL_OP_NO_TLSv1_2 and the other one is. I duplicated the code that was done for other options, so it is consistent with that (I believe). We process the option from the config file, then when we are try to use it we see if SSL actually has the option available in the include file. Not all platforms so. Not all platforms have SSL_OP_NO_TLSv1_2 flag in the SSL header file either which is why this is a duplicate of that code.

paulcallen:
I added a warning if no compression flag is available, although there may be a possibility that compression is not even available so this is irrelevant!
The code is shared for client and server so I don't want to fail the function. I think this is the best we can get without it becoming an odd change.

PR #295, opened 2017-04-25
Kerberos Client and server Authentication.

yakman2020:
Kerberos Encrypt still in progress
@Microsoft/omi-devs

### DIFF
--- a/Unix/configure
+++ b/Unix/configure
@@ -1383,12 +1383,14 @@ fi
 ##
 ##==============================================================================
 
+use_heimdal=0
 case $os in
         LINUX)
             if [ disable_auth != 1 ]
             then
                 gsslib=gssapi_krb5
                 gsslibpath=libgssapi_krb5.so
+                krblibpath=libkrb5.so
             fi
             ;;
         HP-UX | HPUX)
@@ -1396,6 +1398,7 @@ case $os in
             then
                 gsslib=gss
                 gsslibpath=libgss
+                krblibpath=libkrb5
             fi
             ;;
         AIX)
@@ -1403,6 +1406,7 @@ case $os in
             then
                 gsslib=gssapi_krb5
                 gsslibpath="/usr/lib/libgssapi_krb5.a(libgssapi_krb5.a.so)"
+                krblibpath="/usr/lib/libkrb5.a(libkrb5.a.so)"
             fi
             ;;
         SUNOS)
@@ -1410,13 +1414,16 @@ case $os in
             then
                 gsslib=gss
                 gsslibpath=libgss.so
+                krblibpath=libkrb5.so
             fi
             ;;
         DARWIN)
             if [ disable_auth != 1 ]
             then
-                gsslib=/System/Frameworks/GSS.framework/GSS
-                gsslibpath=/System/Frameworks/GSS.framework/GSS
+                use_heimdal=1
+                gsslib=/System/Library/Frameworks/GSS.framework/GSS
+                gsslibpath=/System/Library/Frameworks/GSS.framework/GSS
+                krblibpath=/System/Library/Frameworks/Kerberos.framework/Kerberos
             fi
             ;;
         *)
@@ -1773,6 +1780,7 @@ cat > $fn <<EOF
 #define CONFIG_TMPDIR "$tmpdir"
 #define CONFIG_CIMSCHEMA "$cimschema"
 #define CONFIG_GSSLIB "$gsslibpath"
+#define CONFIG_KRB5LIB "$krblibpath"
 #define CONFIG_SUPPORTS_NTLM $ntlm_supported
 EOF
 
--- a/Unix/http/http.c
+++ b/Unix/http/http.c
@@ -734,25 +734,51 @@ static MI_Boolean _RequestCallbackRead(
 }
 
 /* length of longest description - has to be updated if descriptions are updated */
-#define HTTP_LONGEST_ERROR_DESCRIPTION 50
-static const char* _GetHttpErrorCodeDescription(
-    int httpErrorCode )
+static const char* _HttpErrorCodeDescription(
+    int httpErrorCode, int *pdesclen )
 {
+    static const char STATUS_OK[] = "OK";
+    static const char STATUS_BAD_REQUEST[] = "Bad Request";
+    static const char STATUS_UNAUTHORIZED[] = "Unauthorized";
+    static const char STATUS_INTERNAL_SERVER_ERROR[] = "Internal Server Error";
+    static const char STATUS_UNKNOWN_ERROR[] = "Error";
+
     switch (httpErrorCode)
     {
     case 200:
-        return "OK";
+        if (pdesclen)
+        {
+            *pdesclen = MI_COUNT(STATUS_OK)-1;
+        }
+        return STATUS_OK;
 
     case 400:
-        return "Bad Request";
+        if (pdesclen)
+        {
+            *pdesclen = MI_COUNT(STATUS_BAD_REQUEST)-1;
+        }
+        return STATUS_BAD_REQUEST;
 
     case 401:
-        return "Unauthorized";
+        if (pdesclen)
+        {
+            *pdesclen = MI_COUNT(STATUS_UNAUTHORIZED)-1;
+        }
+        return STATUS_UNAUTHORIZED;
 
     case 500:
-        return "Internal Server Error";
+        if (pdesclen)
+        {
+            *pdesclen = MI_COUNT(STATUS_INTERNAL_SERVER_ERROR)-1;
+        }
+        return STATUS_INTERNAL_SERVER_ERROR;
     }
-    return "Error";
+
+    if (pdesclen)
+    {
+        *pdesclen = MI_COUNT(STATUS_UNKNOWN_ERROR)-1;
+    }
+    return STATUS_UNKNOWN_ERROR;
 }
 
 /*
@@ -767,6 +793,13 @@ static void _ResetWriteState(
         PAL_Free(socketData->sendPage);
         socketData->sendPage = 0;
     }
+
+    if (socketData->sendHeader)
+    {
+        PAL_Free(socketData->sendHeader);
+        socketData->sendHeader = 0;
+    }
+
     socketData->httpErrorCode = 0;
     socketData->authFailed     = FALSE;
     socketData->sentSize = 0;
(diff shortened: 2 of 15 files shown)


yakman2020:
@Microsoft/omi-devs 
Passed pbuild

Unix/pal/host.c
```diff
@@ -0,0 +1,90 @@
...
+    hostname[1023] = '\0';
+
+    gethostname(hostname, 1023);
+
+    memset(&hints, 0, sizeof hints);
+    hints.ai_family = AF_UNSPEC;    /*either IPV4 or IPV6 */
+    hints.ai_socktype = SOCK_STREAM;
+    hints.ai_flags = AI_CANONNAME;
+
+    if ((gai_result = getaddrinfo(hostname, NULL, &hints, &info)) != 0)
```
jeffaco: A lot of this is protocol stuff that I'm not familiar with, so I'll leave that to Paul.

Here, info appears to be something that should be freed via `freeaddrinfo`, is this done?

Unix/configure
```diff
@@ -1383,40 +1383,47 @@ fi
 ##
 ##==============================================================================
 
+use_heimdal = 0
```
paulcallen: I don't see this being used anywhere

Unix/http/http.c
```diff
@@ -816,13 +817,20 @@ static Http_CallbackResult _WriteHeader(
...
+                                         (int)handler->httpErrorCode, _GetHttpErrorCodeDescription(handler->httpErrorCode),
+                                         (int)handler->sendPage->u.s.size, handler->pSendAuthHeader );
+            PAL_Free(handler->pSendAuthHeader);
+            handler->pSendAuthHeader = NULL;
+        }
+        else 
+        {
+            buf_size = (size_t)Snprintf( currentLine, sizeof(currentLineBuff), RESPONSE_HEADER_FMT,
+                                         (int)handler->httpErrorCode, _GetHttpErrorCodeDescription(handler->httpErrorCode),
+                                         (int)handler->sendPage->u.s.size, "" );
```
paulcallen: isn't this going to add an extra blank line to the headers, which actually means we have a blank line at the beginning of the body which is going to mean the content length is going to be off by 2?
Maybe you need two versions of the RESPONSE_HEADER_FMT.
As a secondary note, sprint is one of the slowest functions known to developers. Only sleep is slower :) Maybe we should use memcpy where possible.

Unix/http/httpauth.c
```diff
@@ -1431,20 +1431,36 @@ Converts the SPNEGO authorization header string to an opaque gss token
...
+        return -1;
+    }
+
+    token->value = (unsigned char *)PAL_Malloc(strlen(auth));
     if (!token->value)
     {
         return -1;
     }
-    memset(token->value, 0, strlen(spnegoToken));
+    memset(token->value, 0, strlen(auth));
```
paulcallen: Stop strlen-ing the auth string. Do it once! Big perf hit!

Unix/http/httpauth.c
```diff
@@ -1513,30 +1532,40 @@ static unsigned char *_BuildAuthResponse(_In_ const char *pProtocol,
...
     int header_start_len = 0;
+    int header_tail_len  = 0;
 
     struct _EncodeContext encode_context = { 0 };
 
     switch (httpStatus)
     {
+    case 0: // We will add the header info later. Just build a WWW-Authenticate: 
+        header_start = HEADER_START_AUTH_ONLY;
+        header_start_len = strlen(HEADER_START_AUTH_ONLY);
```
paulcallen: don't strlen constant strings! Note, there are other places this is done so fix them also!

Unix/tests/cli/test_cli.cpp
```diff
@@ -259,7 +263,20 @@ static void VerifyEnvironmentVariables()
     sudoPath = std::getenv("SUDO_PATH");
     ntlmFile = std::getenv("NTLM_USER_FILE");
     ntlmDomain = std::getenv("NTLM_DOMAIN");
+    krb5Realm = std::getenv("OMI_KRB_TESTS_REALM");
+    if (!hostFqdn)
+    {
+        hostFqdn = getFullyQualifiedDomainName();
```
paulcallen: looks like this is leaked

Unix/pal/host.h
```diff
@@ -0,0 +1,30 @@
...
+# include <direct.h>
+#else
+# include <sys/types.h>
+# include <sys/stat.h>
+# include <unistd.h>
+#endif
+
+PAL_BEGIN_EXTERNC
+
+const char* getFullyQualifiedDomainName();
```
paulcallen: Functions start with capital letters!

Unix/configure
yakman2020: Will be. I want to be able to config to use heimdal on other archs besides macos.

Unix/tests/cli/test_cli.cpp
```diff
@@ -2149,4 +2175,292 @@ NitsTestWithSetup(TestOMICLI32_UserOption, TestCliSetup)
...
+    if (runKrbTests && startServer && !travisCI)
+    {
+        NitsDisableFaultSim;
+
+        string out;
+        string err;
+        MI_Char buffer[1024];
+
+        Stprintf(buffer, MI_COUNT(buffer),
+                 MI_T("omicli gi --hostname %T --auth Kerberos -u %T@%T -p %T --port %T oop/requestor/test/cpp { MSFT_President Key 1 }"),
```
JumpingYang001: we only use %T@%T here and others using %T\\%T, looks inconsitent even it is a unittest.
I don't know which one is correct.

JumpingYang001:
@paulcallen , does Kerberos will support all Basic auth platforms?

paulcallen:
@JumpingYang001  We hope to support all platforms, eventually. Our initial requirement is Redhat/CentOS 7 platforms, but the next milestone requires OSX. Having all platforms is going to depend on how many issues we have on those other platforms though, and only time will tell there.

Unix/pal/host.c
yakman2020: On line 2374:        freeaddrinfo(info);

Unix/http/httpauth.c
yakman2020: done

Unix/pal/host.c
```diff
@@ -0,0 +1,90 @@
...
+
+/*
+**==============================================================================
+**
+** Host name (Windows)
+**
+**==============================================================================
+*/
+
+#if defined(_MSC_VER)
```
paulcallen: lets not add this! I am not sure we will be supporting Windows again, and if we do they will get a link error anyway.

Unix/pal/host.h
yakman2020: fixed

Unix/tests/cli/test_cli.cpp
yakman2020: added to StopServerSudo

Unix/http/http.c
yakman2020: Nope. The crlf is in the applied string. 

However, this code has been replaced with more legible code that eliminates the sprintfs

Unix/http/httpauth.c
yakman2020: Removed the sprintf and extra strlens.

yakman2020:
@Microsoft/omi-devs  Updated the commit to address feedback. and rebased. pbuild run.

Unix/http/httpauth.c
yakman2020: Done

Unix/http/http.c
```diff
@@ -814,76 +861,125 @@ static Http_CallbackResult _WriteHeader(
...
+    if (handler->pSendAuthHeader)
     {
-#if ENCRYPT_DECRYPT
-        Page *pOldPage = handler->sendPage;
+        int auth_header_len = strlen(handler->pSendAuthHeader);
 
-        if (!Http_EncryptData(handler, (char**)&currentLine, &buf_size,  &handler->sendPage) ) {
-             
-            // If we fail it was an error. Not encrypting counts as failure Complain and bail
+        //
```
JumpingYang001: Looks like you want to comment something here for //?

Unix/tools/auth_tests_setup.sh
```diff
@@ -94,7 +94,36 @@ if [ "x${username}" != "x" -a "x${userpasswd}" != "x" ]; then
...
+    if [ "$OMI_KRB_TESTS_ENABLED" = "FALSE" ] ; then
+        echo "Kerberos Tests Disabled via OMI_KRB_TESTS_ENABLED"
+        OMI_KRB_RUN_TESTS="false"
+        export OMI_KRB_RUN_TESTS
+    else
+        unset OMI_KRB_RUN_TESTS
+        if klist -s ; then
+           # There is a ticket granting ticket in the credential cache. 
+           # We expect that kerberos has been set up but we need to see if the user 
+           # exists. However, there aren't many good ways to do that. So for now if the
```
JumpingYang001: comment incomplete? something after "So for now if the"?

Unix/http/httpclient.c
```diff
@@ -1917,10 +1919,17 @@ Page* _CreateHttpHeader(
     int r;
     char* p;
 
-#define HTTP_HEADER_FORMAT "%s %s HTTP/1.1\r\n" \
-    "Content-Length: %d\r\n"\
-    "Connection: Keep-Alive\r\n" 
+    static const char HTTP_PROTOCOL_HEADER[]     = "HTTP/1.1";
+    static const char HTTP_PROTOCOL_HEADER_LEN = MI_COUNT(HTTP_PROTOCOL_HEADER)-1;
+
+    static const char CONTENT_LENGTH_HEADER[]     = "Content-Length: ";
```
JumpingYang001: I think we can store string in https://github.com/Microsoft/omi/blob/master/Unix/http/httpcommon.h#L99 as a macro instead of duplicated in http/http.c and http/httpclient.c, just a suggestion.

yakman2020:
@Microsoft/omi-devs  Updated. pls re-review

Unix/pal/host.c
yakman2020: removed

Unix/tests/cli/test_cli.cpp
yakman2020: They are both correct

Unix/tools/auth_tests_setup.sh
yakman2020: good point. Fixed.

Unix/http/httpclient.c
yakman2020: We need to actually have storage to point at.  If the client and host strings were always identical, which they are not, we could move their definition to a single files somewhere as a global. Then we wouldn't know who was using them.  

If we just have them in a macro, we would have more copies of them in the object file, plus we would not be able to see them in the debugger.

Issue #292, opened 2017-04-18
run “noop” operation with "-synchronous" option should return "got noop response"


Wayenyin:
**Good result:** 
Run command without "-synchronous": 
```
/opt/omi/bin/omicli noop
    got noop response
```

**Issue:**
Run command with "-synchronous": 
```
/opt/omi/bin/omicli noop -synchronous
```

Build：omi-1.2.0-76.x86_64

Actual: there is nothing output
Expect: output should be "got noop response"

paulcallen:
@palladia can you take a look.

palladia:
Fixed with #296

PR #296, opened 2017-04-25
Issue #292: omicli noop -synchronous doesn't generate output

palladia:


paulcallen:
* you got the issue number wrong in the commit. I recommend putting that in the description of the pull request instead of in commit message!  That is the recommended way of doing this. If we happen to move (not likely to happen, but!) to a different source control all the issue numbers would relate to something wrong.
* make sure the commit title is less that 70 characters.
* please include @omi-dev in the description when creating the pull request when ready for us to review.

palladia:
@Microsoft/omi-devs

PR #300, opened 2017-05-02
Update Readme to add unixes

paulcallen:
Add location of where to get all platform kits from.
Add Unix supported platforms.

@Microsoft/omi-devs

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,15 +47,28 @@ Linux    | RPM     | x86          | 0.9.8 | [omi-1.2.0-35.ssl_098.x86.rpm](https
 Alternatively, you can now also download from Microsoft Repo. Instructions
 on setting this up can be found [here](https://technet.microsoft.com/en-us/windows-server-docs/compute/Linux-Package-Repository-for-Microsoft-Software).  Follow the instructions for your platform.  You can then use your platform's package tool to install OMI (i.e. "sudo apt-get install omi", or "sudo yum install omi").
 
+The latest version of all supported platform packages can also be found [here](https://github.com/Microsoft/omi-kits), but where possible we recommend downloading from the Microsoft Repo.
+
 ### Supported Linux Operating Systems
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,15 +47,23 @@ Linux    | RPM     | x86          | 0.9.8 | [omi-1.2.0-35.ssl_098.x86.rpm](https
 Alternatively, you can now also download from Microsoft Repo. Instructions
 on setting this up can be found [here](https://technet.microsoft.com/en-us/windows-server-docs/compute/Linux-Package-Repository-for-Microsoft-Software).  Follow the instructions for your platform.  You can then use your platform's package tool to install OMI (i.e. "sudo apt-get install omi", or "sudo yum install omi").
 
+The latest version of all supported platform packages can also be found [here](https://github.com/Microsoft/omi-kits), but where possible we recommend downloading from the Microsoft Repo.
+
 ### Supported Linux Operating Systems
 
 - CentOS Linux 5,6, and 7 (x64)
```
jeffaco: This is mostly incorrect. We support both x86 and x64 for all Linux platforms. From your statement, I'd think we support only 64-bit platforms, and that's not the case.

README.md
```diff
@@ -47,15 +47,23 @@ Linux    | RPM     | x86          | 0.9.8 | [omi-1.2.0-35.ssl_098.x86.rpm](https
...
 - Debian GNU/Linux 6, 7, and 8 (x64)
 - Oracle Linux 5,6, and 7 (x64)
-- Red Hat Enterprise Linux Server 5,6 and 7 (x64)
+- Red Hat Enterprise Linux Server 5 and 6 (x64), and 7 (x64 and PPC)
 - SUSE Linux Enteprise Server 11 and 12 (x64)
```
jeffaco: SuSE 10 is a supported platform as well. In fact, that's actually our build platform!

PR #303, opened 2017-05-03
Ported some changes from httpclientcxx 1.0.6.5 for socket cleanup.

johnkord:
Porting some changes from the old httpclientcxx (omi version 1.0.6.5) that are used to close sockets. The "Cancel" function posts a NotifyItem to perform a "DELETE_HTTP" action, and used to exist in the old httpclientcxx file. I verified that this change (along with some assocaited changes in CM to call this function) resolves the socket leak issue.

### DIFF
--- a/Unix/deprecated/httpclientcxx/httpclientcxx.cpp
+++ b/Unix/deprecated/httpclientcxx/httpclientcxx.cpp
@@ -34,6 +34,8 @@ static sem_t            s_connectionSemaphore;
 static bool             s_connectionSemaphoreInitialized = false;
 static pthread_mutex_t  s_connectionSemaphoreLock;
 static sem_t            s_sendSemaphore;
+static sem_t            s_deleteSemaphore;
+static pthread_mutex_t  s_deleteSemaphoreLock;
 static pthread_mutex_t  s_sendLock;
 static bool             s_notifySet = false;
 
@@ -568,6 +570,7 @@ void IOThread::DeleteHttpTh(NotifyItem* item)
     item->_rep->_httpClient = NULL;
     item->_rep->_destroyed = true;
 
+    sem_post(&s_deleteSemaphore);
     LOGD2((ZT("_DeleteHttpTh - Done")));
 }
 
@@ -651,7 +654,9 @@ Result HttpClient::Connect(
         s_connectionSemaphoreInitialized = true;
         sem_init(&s_connectionSemaphore, 0, 0);
         sem_init(&s_sendSemaphore, 0, 0);
+        sem_init(&s_deleteSemaphore, 0, 0);
         pthread_mutex_init(&s_connectionSemaphoreLock, NULL);
+        pthread_mutex_init(&s_deleteSemaphoreLock, NULL);
         pthread_mutex_init(&s_sendLock, NULL);
     }
 
@@ -724,6 +729,23 @@ Result HttpClient::Connect(
 
     LOGD2((ZT("HttpClient::Connect - OK")));
     return httpclient::OKAY;
+}
+
+void HttpClient::Cancel( void )
+{
+    pthread_mutex_lock(&s_deleteSemaphoreLock);
+
+    NotifyItem* item = new NotifyItem(_rep);
+
+    if (!_rep->_th->PostItem(item))
+    {
+        LOGE2((ZT("HttpClient::Connect - PostItem failed")));
+        pthread_mutex_unlock(&s_deleteSemaphoreLock);
+        return;
+    }
+
+    sem_wait(&s_deleteSemaphore);
+    pthread_mutex_unlock(&s_deleteSemaphoreLock);
 }
 
 Result HttpClient::StartRequest(


johnkord:
Associated with https://github.com/Microsoft/SCXcm/pull/42

johnkord:
Should this be merged to master, and then merged to stable? I'm not familiar with what is appropriate in this case.

jeffaco:
Stable gets updated when testing passes. Just merge to master, when we test for the next release it'll be included into stable.

Unix/deprecated/httpclientcxx/httpclientcxx.cpp
```diff
@@ -726,6 +731,22 @@ Result HttpClient::Connect(
     return httpclient::OKAY;
 }
 
+void HttpClient::Cancel( void )
+{
+    pthread_mutex_lock(&s_deleteSemaphoreLock);
```
jeffaco: The initialization for the semaphore (and the lock) only happens on connect (line 641 above, in HttpClient::Connect). Will this ONLY be called after HttpClient::Connect() was called? If not, then these things won't be set up, and bad things would happen ...

johnkord: Well, I didn't design this code, this is ported from 1.0.6.5, and I don't have knowledge into how NotifyItem and PostItem work at all. But, it appears we do assume that this will get called only after connect is called. I believe StartRequest also makes the same assumption, that it should get called only after a Connect is called (It too also uses semaphores and mutexes that are initialized in Connect).

JumpingYang001:
Travis CI failed on Mac, BTW, Unix/deprecated/httpclientcxx/httpclientcxx.cpp looks like is deprecated, are we still using it?

jeffaco:
@JumpingYang001 Yeah, this code is still used by the SCXcm project, according to @johnkord. I spoke with him about this yesterday.

JumpingYang001:
@jeffaco , so cm now used old httpclient lib, it is a little different with provided OMI standalone package's omi, right? I mean they have their own omi code which is different with official omi, right? I am not cm guys so I don't have permission to access it and don't know much about it. :)

jeffaco:
CM doesn't have their own OMI code, as such, they just used the depreciated HTTP client code. That's my understanding, anyway.

To be honest, I don't know much about CM either - it's not my project. That said, I was happy to give you read access to the CM repo. You have it now. Follow up in E-Mail if you have problems/questions with access.

JumpingYang001:
@jeffaco , thanks for giving me access, they are actually built deprecated folder here: https://github.com/Microsoft/SCXcm/blob/0344d86549246ae7f4a89ee6ee90b7eb5cd04688/Unix/shared/omi/OMI_SSL-0.9.8.mak#L12.

Depreciated HttpClient is a class here:
https://github.com/Microsoft/omi/blob/15d9c277faa32511c9a2ca7ab20efbbaa93c536c/Unix/deprecated/httpclientcxx/httpclientcxx.h#L150

Default HttpClient is a struct here:
https://github.com/Microsoft/omi/blob/89ebdeab20cfb2d47249d4c73d27f603aef3b588/Unix/http/httpclient_private.h#L111
```
struct _HttpClient {
    MI_Uint32 magic;
    Selector internalSelector;
    Selector *selector;
    HttpClientCallbackOnConnect callbackOnConnect;
    HttpClientCallbackOnStatus2 callbackOnStatus;
    HttpClientCallbackOnResponse callbackOnResponse;
    void *callbackData;
    SSL_CTX *sslContext;

    HttpClient_SR_SocketData *connector;

    MI_Boolean internalSelectorUsed;
};
```

https://github.com/Microsoft/omi/blob/5e4f91083d9129be55b466192e4981e843a7778f/Unix/http/httpclient.h#L23
```
typedef struct _HttpClient HttpClient;
```

JumpingYang001:
@johnkord , CI failed on Mac due to running more than 10 minutes, maybe you can re-trigger the failed jobs.

johnkord:
@Microsoft/omi-devs 
I retriggered the Mac OS build, and everything succeeded.  These changes have no effect on an OMI build, since OMI does not build the deprecated folder by default.  SCXCM will build this file and include the resultant library with its package, so this change shouldn't require a new OMI release.  I believe I do not have the proper permissions to merge this to master, so I need someone with the right permissions to do so.

jeffaco:
@johnkord OSTC-devs has write access to OMI, which means that you can merge this directly.

However, I noted I'm the only reviewer here. Just for sanity sake, can you ask @paulcallen to review this as well? Once he reviews, you should be able to commit on your own, let me know if you have any problems.

Be sure to rebase on master, your branch is behind, before the merge.

Unix/deprecated/httpclientcxx/httpclientcxx.cpp
```diff
@@ -726,6 +731,22 @@ Result HttpClient::Connect(
...
+
+    NotifyItem* item = new NotifyItem(_rep);
+
+    if (!_rep->_th->PostItem(item))
+    {
+        LOGE2((ZT("HttpClient::Connect - PostItem failed")));
```
paulcallen: you did not 
    pthread_mutex_unlock(&s_deleteSemaphoreLock); 
in failure case. It Does the PostItem always delete the NotifyItem even in failure case, if not then you need to delete also,

johnkord: Thanks Paul, nice catch. I've fixed this now. This is also a bug in 1.0.6.5.

johnkord:
@paulcallen This now contains the change you requested and is ready for re-review

PR #304, opened 2017-05-08
Compiles successfully on armv7 (raspberry pi).

yakman2020:
@Microsoft/omi-devs a Runs regress successfully. Need to pbuild before merge but don't expect any problems.

### DIFF
--- a/Unix/buildtool
+++ b/Unix/buildtool
@@ -272,6 +272,13 @@ if [ -z "$target" ]; then
             distro=`lsb_release -i | awk -F":" '{ print $2 }'`
             distro_version=`lsb_release -r | awk -F":" '{ print $2 }'`
 	    ;;
+        arm*:Linux:*)
+            os=LINUX
+            arch=ARM
+            compiler=GNU
+            distro=`lsb_release -i | awk -F":" '{ print $2 }'`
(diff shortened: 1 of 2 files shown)


JumpingYang001:
OMI will support mobile phone, people will use OMI as a client or server in their phone?

yakman2020:
In order to support mobile we would have to have an ARM package for android. This just makes it work on ARM hosts running linux.  That is interesting because there are a bunch of IoT hosts running linux on the arm.

Anyway, just one more architecture.

yakman2020:
@Microsoft/omi-devs  Passed pbuild

Unix/buildtool
```diff
@@ -966,7 +976,7 @@ if [ "$arg1" = "cshlibflags" -o "$arg1" = "cxxshlibflags" ]; then
     r=""
 
     case "$platform" in
-        LINUX_IX86_GNU|LINUX_X86_64_GNU|LINUX_PPC_GNU)
+        LINUX_IX86_GNU|LINUX_X86_64_GNU|LINUX_PPC_GNU|LINUX_ARM_GNU)
```
jeffaco: These are duplicated so many times, I wonder if you'd be better off defining a GENERIC_LINUX symbol earlier and then just testing that. Then, in the future, if we need to add a platform, we add it in one place?

PR #308, opened 2017-05-10
Create setup-kerberos-omi.md

yakman2020:


Unix/doc/setup-kerberos-omi.md
```diff
@@ -0,0 +1,183 @@
+#Enabling Kerberos Authentication on OMI
```
JumpingYang001: Need add a blank after all # or ##, so the markdown format can show correctly, thanks.

PR #314, opened 2017-05-18
Enable some providers for testing

JumpingYang001:
@Microsoft/omi-devs , I have enabled some providers for testing: FileSystem, alert, TestClass_AllDMTFArrays, could you help to review this PR? thanks very much!

We only need these providers on Linux/Unix platforms, thanks.


JumpingYang001:
Mac failed with below error, how can I disable building it on Mac? We don't need any providers on Mac.
```
clang -c -o /Users/travis/build/Microsoft/omi/Unix/output1/obj/samples/Providers/Indication/util/util.o -g -fPIC -Werror -Wall -I/usr/local/include -Dmacos -fvisibility=hidden -fstack-protector-all -I/usr/local/opt/openssl/include -I/Users/travis/build/Microsoft/omi/Unix/output1/include -I../../../.. -I../../../../common -DMI_CONST= -DHOOK_BUILD -DMI_API_VERSION=2 util.c 

[1mutil.c:291:12: [0m[0;1;31merror: [0m[1m'sem_init' is deprecated [-Werror,-Wdeprecated-declarations][0m

    return sem_init(self->sem, 0, count) == 0 ? 0 : -1;

[0;1;32m     
```

JumpingYang001:
I have fixed the Mac build issue, but 1 case failed on Mac:
```
libtest_cli.dylib:TestOMICLI28_FaultWsman...
test_cli.cpp(2054): 
	Compare: Exec(buffer, out, err) == 5
		LHS: 1
		RHS: 5
		(Omicli error)
test_cli.cpp(2058): 
	Compare: err.c_str() == expect.c_str()
		LHS: omicli: result: MI_RESULT_FAILED
omicli: result: Could not connect

		RHS: omicli: result: MI_RESULT_INVALID_CLASS
omicli: result: The specified class does not exist.

		(Error mismatch)
test_cli.cpp(2059): 
	Compare: out == "" == true
		LHS: 0
		RHS: 1
		(Output mismatch)
	[Failed]  libtest_cli.dylib:TestOMICLI28_FaultWsman	0.121914

```

JumpingYang001:
Re-trigger to check if the Mac fail case works.

Unix/samples/Providers/Indication/util/util.h
```diff
@@ -221,8 +225,12 @@ typedef struct _Sem
 #if defined(_MSC_VER)
     HANDLE handle;
 #else
+#ifdef __APPLE__
```
jeffaco: This would be better as `#elif` (combine 227 and 228, like this):

```
#elif defined(__APPLE__)
```

Unix/samples/Providers/Indication/util/util.h
```diff
@@ -246,7 +254,11 @@ MI_INLINE void Sem_Destroy(
 #else
     if (self->sem)
     {
+#ifdef __APPLE__
+    dispatch_release(self->sem);
```
jeffaco: I'm not really familiar with the `dispatch_release` and related calls on Mac without investigation. Please have @yakman2020 (Bruce) review this, he's probably best.

Unix/samples/Providers/Indication/util/util.h
```diff
@@ -259,8 +271,13 @@ MI_INLINE int Sem_Wait(
 #if defined(_MSC_VER)
     return WaitForSingleObject(self->handle, INFINITE) == WAIT_OBJECT_0 ? 0 : -1;
 #else
+#ifdef __APPLE__
```
jeffaco: Combine 273 and 274, as mentioned above.

Unix/installbuilder/GNUmakefile
```diff
@@ -159,6 +159,13 @@ else #($(PF),Darwin)
 	# copy normal libmi.so for other platforms
 	cp $(OUTPUTDIR)/lib/libmi.so $(PACKAGE_DIR)
   endif #ifeq ($(PF),Linux)
+ 
+  # copy some providers for Linux/Unix platforms
+  mkdir -p $(PACKAGE_DIR)/test
+  cp $(OUTPUTDIR)/lib/libalert.so $(PACKAGE_DIR)/test
+  cp $(OUTPUTDIR)/lib/libfilesys.so $(PACKAGE_DIR)/test
+  cp $(OUTPUTDIR)/lib/libPresidentProvider.so $(PACKAGE_DIR)/test
+  cp $(OUTPUTDIR)/lib/libTestClass_AllDMTFArrays.so $(PACKAGE_DIR)/test
```
jeffaco: At least these are in a subdirectory, that's good. But yuck, that means that every single build, on every single platform, is going to need these. And there are a lot of test providers, this list can grow longer.

Do you really need these providers for each and every platform for each and every build? Remember, you can build these but not stage them. That way, you can just build a specific platform if needed, and it doesn't end up on the build share. It's cleanliness of the build share that is my biggest concern here.

If you REALLY need these for every single platform on every single build, then either we put these in a `test` directory, as you did here, or we put them in some other artifact directory (which Jenkins does support), and copy them to some other location.

Please give guidance of how often you realistically need these, and we'll go from there.

Unix/samples/Providers/Indication/util/util.h
paulcallen: please use pal/sem.h Sem structure and the associated semaphore implementation in there instead. That does all the platform specific stuff needed to implement semaphore on different platforms.

Unix/installbuilder/GNUmakefile
JumpingYang001: @jeffaco , we only need build **one time** to get the .so files, since we need them on every Linux/Unix platform, could you help to build them to a folder that I can use them? thanks very much!
I can close this PR if you can help me build out lib libalert.so,libfilesys.so,libPresidentProvider.so,libTestClass_AllDMTFArrays.so on every platform, thanks. I only need build one time to get the libs, thank you!

JumpingYang001:
@paulcallen , I fixed to remove old Sem and use pal/sem.h, thanks.
@yakman2020 , I don't have mac to build.
@jeffaco , do I need to revert Unix/installbuilder/GNUmakefile and you help me build to a folder then we can use it? thank you!

jeffaco:
@JumpingYang001 You do have access to a Mac, it's called the shared DEV systems. All of us run [pbuild](https://github.com/microsoft/pbuild) prior to checkin. For OMI, that includes builds on the Mac. When you got write access to the OMI repo, it was with the understanding that you follow the same checkin rules that we follow. That includes a successful PBUILD on the Mac platform for OMI.

Additionally, if you run PBUILD, you will get your shared libraries for the providers across all of our platforms. You should keep your Makefile changes to build the providers but revert the Makefile changes to stage them to the build shares. If you run PBUILD (and you must in order to check in), then each of the PBUILD machines will build the shared libraries as part of their build. You can copy them off and then you're all set.

I'm not signing off yet; I'd like the above to be handled first, thanks.

JumpingYang001:
@jeffaco , I will build the providers using PBuild next week, and then update current commit to exclude makefiles with "Fix sem part for Indication provider", thanks.

jeffaco:
To disable PBUILD on the Mac, running pbuild.py with --help is your friend. You'll note:

```
  --exclude=EXCLUDE     Overrides default exclude list from configuration file
                        (if any); comma-separated list of hosts to exclude
                        from the build
```

That's a very simple way to disable a single host (or a list of hosts) from a build. Sorry I missed this question earlier.

Unix/samples/Providers/TestClass_AllDMTFArrays/TestClass_AllDMTFArrays.cpp
```diff
@@ -119,14 +120,15 @@ void SetUint32Array(TestClass_AllDMTFArrays *instance, MI_Uint32 count)
     {
         for(MI_Uint32 i=0; i < count; i++)
         {
-            uint32Array[i] =  (MI_Uint32) i % 4294967296;
+            uint32Array[i] =  (MI_Uint32) i % 0xFFFFFFFF;
```
jeffaco: Include file <limits.h> can be helpful here. You'd have `LONG_MAX`, `ULONG_MAX`, etc.

The problem with the original code is you didn't have `L` or `UL` decorations for data types of the constants. The definitions in <limits.h> have that.

This is fine is as, just pointing it out. I think the constants in <limits.h> would be more clear, but I won't hold up this review just for that.

Unix/samples/Providers/Indication/Lifecycle/XYZ_ComputerSystem.c
```diff
@@ -10,6 +10,7 @@
 /* @migen@ */
 #include <MI.h>
 #include <util.h>
+#include <pal/sem.h>
```
jeffaco: I don't fully understand, sorry. You're including this include file, but I don't see it referenced in code anywhere (i.e. actually calling it). Am I missing something?

JumpingYang001: The wrapper in pal/sem.h is very good, so we needn't change any method in code, one calling is here: https://github.com/Microsoft/omi/blob/master/Unix/samples/Providers/Indication/Lifecycle/XYZ_ComputerSystem.c#L226

Unix/samples/Providers/TestClass_AllDMTFArrays/TestClass_AllDMTFArrays.cpp
JumpingYang001: Yeah, it will be better: ULONG_MAX->4294967295 , LONG_MAX->2147483647 , LLONG_MAX ->9223372036854775807 , ULLONG_MAX ->18446744073709551615 .

JumpingYang001:
@jeffaco , I have fixed base on you comments, thanks.

jeffaco:
@paulcallen The semaphore changes have been made, please re-review and sign off if appropriate so Jumping can finally commit this.

PR #316, opened 2017-05-19
Fixes kinit when cred cache is reset or expired

yakman2020:
It works well enough we need to make sure solaris and aix machines have the kerberos tests turned off by the environment variable OMI_KRB_TESTS_ENABLED=FALSE in /etc/profile.  I have done this for any machines I use.


yakman2020:
@Microsoft/omi-devs

PR #318, opened 2017-05-30
Add support for query option in omiconfigeditor

vimish:
@Microsoft/ostc-devs 

omiconfigeditor will now support querying for a
given value in a property. If the value is found
then the exit code is 0 and if not then exit code
is 1.

Sample command:
omiconfigeditor httpsport -q 1270 < /etc/opt/omi/conf/omiserver.conf

### DIFF
--- a/Unix/configeditor/configeditor.cpp
+++ b/Unix/configeditor/configeditor.cpp
@@ -44,6 +44,7 @@ OPTIONS:\n\
     -s, --set VALUE             Set property name to specific value.\n\
     -c, --comment               Comment out property name specified.\n\
     -u, --uncomment             Uncomment out property name specified.\n\
+    -q, --query VALUE           Check if property contains queried value.\n\
 \n\
 EXAMPLES:\n\
     The following will add \"1270\" to property httpsport in file\n\
@@ -97,6 +98,7 @@ struct Options
     string add;
     string remove;
     string set;
+    string query;
 
     Options() : debug(false), help(false), comment(false), uncomment(false)
     {
@@ -126,6 +128,8 @@ static void GetCommandLineOptions(
         "--comment",
         "-u",
         "--uncomment",
+        "-q:",
+        "--query:",
         NULL
     };
 
@@ -169,7 +173,8 @@ static void GetCommandLineOptions(
                 options.uncomment ||
                 !options.add.empty() ||
                 !options.remove.empty() ||
-                !options.set.empty())
+                !options.set.empty() ||
+                !options.query.empty())
             {
                 err(ZT("Conflicting option specified with: %s"), scs(state.opt));
             }
@@ -204,6 +209,14 @@ static void GetCommandLineOptions(
                      strcmp(state.opt, "--set") == 0)
             {
                 options.set = state.arg;
+            }
+            else if (strcmp(state.opt, "-q") == 0 || 
+                     strcmp(state.opt, "--query") == 0)
+            {
+                options.query = state.arg;
+
+                if (options.query.find(",") != string::npos)
+                    err(ZT("Not allowed to query element with embedded comma: %s"), scs(state.arg));
             }
             else
             {
@@ -319,7 +332,7 @@ int MI_MAIN_CALL main(int argc, const char** argv)
 
     string arg1(argv[1]);
 
-    if (!opts.comment && !opts.uncomment && opts.add.empty() && opts.remove.empty() && opts.set.empty())
+    if (!opts.comment && !opts.uncomment && opts.add.empty() && opts.remove.empty() && opts.set.empty() && opts.query.empty())
     {
         Ftprintf(stderr, ZT("Usage: %s key [OPTIONS]\n"), scs(arg0));
         Ftprintf(stderr, ZT("One option (an action) must be specified\n"));
@@ -329,6 +342,7 @@ int MI_MAIN_CALL main(int argc, const char** argv)
 
     // Read from stdin, writing results to stdout with modifications as necessary
 
+    bool fDontEchoFile = !opts.query.empty() ? true : false;
     string line;
     while ( getline(cin, line) )
     {
@@ -336,7 +350,10 @@ int MI_MAIN_CALL main(int argc, const char** argv)
         size_t pos = line.find("=");
         if (line.empty() || pos == string::npos)
         {
-            cout << line << std::endl;
+            if (!fDontEchoFile)
+            {
+                cout << line << std::endl;
+            }
             continue;
         }
 
@@ -345,7 +362,10 @@ int MI_MAIN_CALL main(int argc, const char** argv)
         _TrimLeadingTrailingSpaces(prop);
         if (prop.empty())
         {
-            cout << line << std::endl;
+            if (!fDontEchoFile)
+            {
+                cout << line << std::endl;
+            }
             continue;
         }
 
@@ -372,7 +392,10 @@ int MI_MAIN_CALL main(int argc, const char** argv)
         // If property isn't the one we want, move along
         if (prop != arg1)
         {
-            cout << line << std::endl;
+            if (!fDontEchoFile)
+            {
+                cout << line << std::endl;
+            }
             continue;
         }
 
@@ -423,13 +446,30 @@ int MI_MAIN_CALL main(int argc, const char** argv)
             }
             ss << commentByte << prop << "=" << _CreateStringFromList(valueString,valueList);
         }
+        // Support querying a value from a property list
+        else if (!opts.query.empty())
+        {
+        	std::list<std::string> valueList;
+
+        	_ParseValueList(valueList, value);
+        	if (_IsValueInList(valueList, opts.query) != valueList.end())
+        	{
+        	    cout << prop << " have the value " << opts.query << std::endl;
+        	    return 0;
+        	}
+               cout << prop << " does not have the value " << opts.query << std::endl;
+        	return 1;
+        }
         else
         {
             // Um, what are we supposed to do?
             err(ZT("Logic error, file %s, line %d"), scs(__FILE__), __LINE__);
         }
 
-        cout << ss.str() << std::endl;
+        if (!fDontEchoFile)
+        {
+            cout << ss.str() << std::endl;
+        }
     }
 
     if (!cin.eof())


jeffaco:
@omi-devs should review this as well. Note that I was the one that wrote omiconfigeditor, though, so for @omi-devs this is mostly FYI. I can review just fine.

Unix/configeditor/configeditor.cpp
```diff
@@ -301,7 +314,7 @@ int MI_MAIN_CALL main(int argc, const char** argv)
     Options opts;
 
     GetCommandLineOptions(argc, argv, opts);
-
+
```
jeffaco: Looks like you added spurious spaces here, please remove them.

Unix/configeditor/configeditor.cpp
```diff
@@ -336,16 +349,22 @@ int MI_MAIN_CALL main(int argc, const char** argv)
         size_t pos = line.find("=");
         if (line.empty() || pos == string::npos)
         {
-            cout << line << std::endl;
+        	if (opts.query.empty())
+        	{
+                cout << line << std::endl;
```
jeffaco: First, fix indentation here.

Next, this change is necessary because you'd never want to echo lines for a query option, yeah? That makes sense, but if somebody made a mistake and redirected to the new omiserver.conf file, that would effectively lose it.

Not sure what's right here. In one sense, it might make sense to redirect the output anyway and let you pipe it to /dev/null if you don't want it. At least that way, if you inadvertently redirected with query, you wouldn't lose your omiserver.conf file. But on the other hand, we don't normally EVER want the output for a query, so maybe that's okay. I'm just pointing it out; if you like it this way, I'm fine with that.

I think if you keep it make it flag driven (i.e. `fDontEchoFile`). That way, if other options need it down the road, they can just set the flag and you don't have to add `||` conditions here.

Unix/configeditor/configeditor.cpp
```diff
@@ -336,16 +349,22 @@ int MI_MAIN_CALL main(int argc, const char** argv)
...
+        	}
             continue;
         }
-
+
```
jeffaco: Eliminate spurious spaces

Unix/configeditor/configeditor.cpp
```diff
@@ -336,16 +349,22 @@ int MI_MAIN_CALL main(int argc, const char** argv)
...
         }
-
+        
         // Parse the property and value from the line
         string prop = line.substr(0, pos);
         _TrimLeadingTrailingSpaces(prop);
         if (prop.empty())
         {
-            cout << line << std::endl;
+        	if (opts.query.empty())
```
jeffaco: I definitely like the idea of a flag here. Also, please combine the flag with the other condition at line 362. It won't make a code optimization difference, but no reason not to do both conditions in a single `if` statement.

Unix/configeditor/configeditor.cpp
```diff
@@ -336,16 +349,22 @@ int MI_MAIN_CALL main(int argc, const char** argv)
...
+        	{
+                cout << line << std::endl;
```
jeffaco: Indentation issue here, I think. Or is this tab indented and just showing up this way?

Unix/configeditor/configeditor.cpp
```diff
@@ -372,7 +391,10 @@ int MI_MAIN_CALL main(int argc, const char** argv)
         // If property isn't the one we want, move along
         if (prop != arg1)
         {
-            cout << line << std::endl;
+        	if (opts.query.empty())
```
jeffaco: Ditto with above, and combine with 392.

Unix/configeditor/configeditor.cpp
```diff
@@ -423,13 +445,29 @@ int MI_MAIN_CALL main(int argc, const char** argv)
...
+        }
         else
         {
             // Um, what are we supposed to do?
             err(ZT("Logic error, file %s, line %d"), scs(__FILE__), __LINE__);
         }
-
-        cout << ss.str() << std::endl;
+        
+        if (opts.query.empty())
```
jeffaco: Ditto above. And I think 466 has spaces in it, blank lines should not have spaces.

JumpingYang001:
Should be @Microsoft/omi-devs , thanks.

Unix/configeditor/configeditor.cpp
```diff
@@ -44,6 +44,7 @@ OPTIONS:\n\
     -s, --set VALUE             Set property name to specific value.\n\
     -c, --comment               Comment out property name specified.\n\
     -u, --uncomment             Uncomment out property name specified.\n\
+    -q, --query                 Check if property contains queried value.\n\
```
JumpingYang001: to be consistent with previous usage, may below you'd better use below for this:
```
-q, --query VALUE             Check if property contains queried value.\n\ 
```

Unix/configeditor/configeditor.cpp
JumpingYang001: I think we should keep the output no matter the query is specified or not specified.

Unix/configeditor/configeditor.cpp
JumpingYang001: I guess @vimish should use Notepad++, as it default use this format, looks like you should make some change on the setting, I didn't try it, I also use Notepad++ and I deleted the spaces in new lines manually...
for the indent in Notepad++, I have set in setting to make 4 spaces instead of tab.

Unix/configeditor/configeditor.cpp
```diff
@@ -44,6 +44,7 @@ OPTIONS:\n\
     -s, --set VALUE             Set property name to specific value.\n\
     -c, --comment               Comment out property name specified.\n\
     -u, --uncomment             Uncomment out property name specified.\n\
+    -q, --query VALUE           Check if property contains queried value.\n\
```
vimish: Updated the help string

Unix/configeditor/configeditor.cpp
```diff
@@ -372,7 +392,10 @@ int MI_MAIN_CALL main(int argc, const char** argv)
         // If property isn't the one we want, move along
         if (prop != arg1)
         {
-            cout << line << std::endl;
+            if (!fDontEchoFile)
```
vimish: @jeffaco 
Added the flag to determine output to stdout.

I did not chose the current behavior of redirecting from stdin to stdout for following reason:
1. My current implementation exits as soon as the property in query is reached. This would mean that complete omiserver.conf might not get redirected.
2. As of now query op outputs string like **<property> has value <value>** or **<property> does not have value <value>** . With this I intended that someone running the command manually can get to know the result without looking at the exit code.  Hence if stdout is redirected to /dev/null these strings too will not get displayed.

Let me know if this makes sense or I should change my implementation to redirect complete file to stdout for query operation to.

vimish: Also I cannot combine the two if statements as when the 1st if is true then both cout and continue needs to be executed. When the second if is true then cout should not get executed but continue still needs to be executed. I could not think of better way than this, other than possibly using an if, else-if blocks. 
Let me know if an missing something here.

vimish:
@Microsoft/omi-devs

Please review the next iteration.

@jeffaco
I have run PBUILD against the hosts which I use for my scx PBUILD runs. Please let me know if I need to run against any other set of hosts.

Unix/configeditor/configeditor.cpp
jeffaco: If you don't want to output the omiserver.conf file, that's fine.

Outputting the string "has value" or "does not have value" means that callers, at least from postinstall, will need to redirect stdout to /dev/null. That's fine, and is consistent with other Linux utilities (grep, etc). Just pointing this out so you're aware when you start using this new feature.

On combining the `if` statements, brain fart on my part. I was looking/reviewing too quickly. Of course you're right, you can't combine in this case.

Unix/configeditor/configeditor.cpp
```diff
@@ -329,14 +342,18 @@ int MI_MAIN_CALL main(int argc, const char** argv)
 
     // Read from stdin, writing results to stdout with modifications as necessary
 
+    bool fDontEchoFile = opts.query.empty() ? false : true;
```
jeffaco: This could have just been `bool fDontEchoFile = !opts.query.empty()`, which is a bit more readable.

Unix/configeditor/configeditor.cpp
```diff
@@ -423,13 +446,30 @@ int MI_MAIN_CALL main(int argc, const char** argv)
...
+        {
+        	std::list<std::string> valueList;
+        	
+        	_ParseValueList(valueList, value);
+        	if (_IsValueInList(valueList, opts.query) != valueList.end())
+        	{
+        	    cout << prop << " have the value " << opts.query << std::endl;
+        	    return 0;
+        	}
+                cout << prop << " does not have the value " << opts.query << std::endl;
```
jeffaco: Indentation for cout is off by one.

What sort of editor are you using where this sort of thing happens?

Unix/configeditor/configeditor.cpp
```diff
@@ -423,13 +446,30 @@ int MI_MAIN_CALL main(int argc, const char** argv)
             }
             ss << commentByte << prop << "=" << _CreateStringFromList(valueString,valueList);
         }
+        // Support querying a value from a property list
+        else if (!opts.query.empty())
+        {
+        	std::list<std::string> valueList;
+
```
jeffaco: Needless spaces in here. Blank line is fine (and good), prefer blank line without spaces.

Unix/configeditor/configeditor.cpp
```diff
@@ -423,13 +446,30 @@ int MI_MAIN_CALL main(int argc, const char** argv)
...
+        	return 1;
+        }
         else
         {
             // Um, what are we supposed to do?
             err(ZT("Logic error, file %s, line %d"), scs(__FILE__), __LINE__);
         }
-
-        cout << ss.str() << std::endl;
+
```
jeffaco: Needless spaces in here. Blank line is fine (and good), prefer blank line without spaces.

A project I'm working on, mesos, actually as a git commit validity checker that would reject commits that have this sort of thing ...

jeffaco:
Also be sure to rebase with master to allow for FF commit.

PR #323, opened 2017-06-13
crash when failing to connect to remote machine

paulcallen:
Connecting to an invalid machine caused a crash
@Microsoft/omi-devs

### DIFF
--- a/Unix/miapi/InteractionProtocolHandler.c
+++ b/Unix/miapi/InteractionProtocolHandler.c
@@ -522,12 +522,18 @@ static void InteractionProtocolHandler_Operation_Strand_Close( _In_ Strand* self
     if (operation->protocolConnection->type == PROTOCOL_SOCKET)
     {
         ProtocolSocketAndBase* protocol = operation->protocolConnection->protocol.socket;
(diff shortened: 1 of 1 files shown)


Unix/miapi/InteractionProtocolHandler.c
```diff
@@ -1198,18 +1211,14 @@ MI_Result InteractionProtocolHandler_Session_Connect(
...
             {
                 /* This can happen if the protocol operation failed and already posted a result to the client */
                 trace_MI_SocketConnectorFailed(operation, r);
-                if (operation->parentSession)
-                {
-                    SessionCloseCompletion_Release(operation->parentSession->sessionCloseCompletion);
-                }
+                PAL_Free(operation->protocolConnection);
+                operation->protocolConnection = NULL;
                 goto done;
```
palladia: This case would return MI_RESULT_OK.  This is what we want?

paulcallen: correct. We need to return TRUE from here because the error is processed in the protocol handler itself and posted via the strand

PR #331, opened 2017-06-19
Kerberos Authentication and test changes required by MacOS

yakman2020:
@Microsoft/omi-devs 

Pbuild completed.

### DIFF
--- a/Unix/http/httpclientauth.c
+++ b/Unix/http/httpclientauth.c
@@ -42,6 +42,15 @@
 #include "httpkrb5.h"
 #endif
 
+#if !defined(GSS_IO_BUFFER_TYPE_PADDING)
+#define GSS_IOV_BUFFER_TYPE_PADDING 9 // For the mac, this is normally in the header file
+                                      // /System/Library/Frameworks/GSS.framework/Headers/gssapi.h which is included in the 
+                                      // GSS framework headers directory. However, its not coming through, and when it is included
+                                      // directly, there are a large number of duplicate declarations. Its one number, and its
+                                      //  not going to change so I just define it here. For linux, it is located in /usr/include/gssapi/gssapi_ext.h
+                                      //  which we don't include because it isn't present on some platforms.  Same logic
+#endif
+
 #ifdef CONFIG_POSIX
 #include <openssl/ssl.h>
 #include <openssl/err.h>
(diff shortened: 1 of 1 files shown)


Unix/http/httpclientauth.c
```diff
@@ -1436,23 +1461,22 @@ HttpClient_EncryptData(_In_ HttpClient_SR_SocketData * handler, _Out_ Page **pHe
         iov[1].type = GSS_IOV_BUFFER_TYPE_DATA;
         iov[1].buffer.length = original_content_len;
         iov[1].buffer.value  = poriginal_data;
+        iov[2].type = 9 /*GSS_IOV_BUFFER_TYPE_PADDING*/  | GSS_IOV_BUFFER_FLAG_ALLOCATE;
```
paulcallen: how does this build?

Unix/http/httpclientauth.c
```diff
@@ -728,6 +729,28 @@ static _Success_(return == 0) int _GssClientInitLibrary( _In_ void* data, _Outpt
         _g_gssClientState.Gss_Wrap   = (Gss_Wrap_Func) fn_handle;
 
 #if GSS_USE_IOV
+#if defined(macos)
```
jeffaco: Please combine 731 and 732 into a single line.

Unix/http/httpclientauth.c
```diff
@@ -728,6 +729,28 @@ static _Success_(return == 0) int _GssClientInitLibrary( _In_ void* data, _Outpt
...
+        fn_handle = dlsym(libhandle, "__ApplePrivate_gss_unwrap_iov");
```
jeffaco: Mac is an interesting case in that it's a single platform with far less variability than other platforms. For example, you don't have a choice as to what libraries are installed; these tend to just come with the system.

Given this, would it make sense to statically link against MacOS libraries, or dynamically link against them. I'm not saying what you're doing is wrong. I just want you to consider if this is the right approach or not. To be honest: I'm not sure. But I did want you to consider it.

Unix/http/httpclientauth.c
jeffaco: I'm not sure how this would not build unless `GSS_IOV_BUFFER_FLAG_ALLOCATE` was not defined. Type 9 (presumably `GSS_IOV_BUFFER_TYPE_PADDING`, which must not be defined or Bruce would have used it) ored with `GSS_IOV_BUFFER_FLAG_ALLOCATE`.

What am I missing?

Unix/http/httpclientauth.c
jeffaco: Oops, there's a `#else`, scratch this!

Unix/http/httpclientauth.c
paulcallen: Fine! I didn't see the 9, just the commented out value!
Bruce is going to define that value in the header so it will be all nice again :)

paulcallen: I was confused! Didn't see the 9. Bruce will add a #define to the header

Unix/http/httpclientauth.c
paulcallen: I prefer the fact it is the same on all platforms in terms of dlsym.

Unix/http/httpclientauth.c
paulcallen: you added the constant in the header but aren't using it

yakman2020: iov[0].type = GSS_IOV_BUFFER_TYPE_HEADER     | GSS_IOV_BUFFER_FLAG_ALLOCATE;
        iov[1].type = GSS_IOV_BUFFER_TYPE_DATA;
        iov[1].buffer.length = original_content_len;
        iov[1].buffer.value  = poriginal_data;
        iov[2].type = GSS_IOV_BUFFER_TYPE_PADDING  | GSS_IOV_BUFFER_FLAG_ALLOCATE;

PR #332, opened 2017-06-20
Fixes problems with PAM auth token corner cases on solaris

yakman2020:
Greatly decrease problems with pbuild auth failures.

### DIFF
--- a/Unix/scripts/installpam-sun9
+++ b/Unix/scripts/installpam-sun9
@@ -9,7 +9,8 @@ get_new_pam_config_file() {
     # No match found
     # sshd not explicitly configured.
     # Use passwd
-    sshd_conf=`echo "omi    auth requisite          pam_authtok_get.so.1
+    sshd_conf=`echo "auth       optional       pam_krb5.so use_first_pass use_kcminit default_principal
+omi    auth requisite          pam_authtok_get.so.1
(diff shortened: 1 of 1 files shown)


yakman2020:
@Microsoft/omi-devs  pls review

Unix/scripts/installpam-sun9
```diff
@@ -9,9 +9,11 @@ get_new_pam_config_file() {
     # No match found
     # sshd not explicitly configured.
     # Use passwd
-    sshd_conf=`echo "omi    auth requisite          pam_authtok_get.so.1
+    sshd_conf=`echo "auth       optional       pam_krb5.so use_first_pass use_kcminit default_principal
```
jeffaco: First, you're referencing a new PAM file. Are you **certain** that that PAM file exists on the target in all cases?

Second, this file is installpam_sun9, and Solaris 9 isn't actually a supported platform for 2016 products. This means that OMI is no longer built on Solaris 9 (it's long retired). There is a sun10 file that is used for Solaris 10 and later.

Given this, I'm pretty certain this should be closed without committing ... if want, in a future commit, you should be able to `git rm` the `installpam_sun9` file, pbuild it, and commit that ...

yakman2020: There it is... OK

PR #333, opened 2017-06-20
Adding some features to support the OMI-Script-Provider.

EMumau:
The changes to provreg and provmgr add support for two new .reg file members,
interpreter and startup, which are used for identifying and starting providers
that use the OMI-Script-Provider.
The changes to omireg enable the omireg tool to read additional commandline
arguments from a configuration file to support registering providers for the
OMI-Script-Provider.  The use of the configuration file allows for future
scripting languages to be supprted without changing the OMI code.


Unix/provmgr/provmgr.c
```diff
@@ -165,16 +165,22 @@ static Library* MI_CALL _OpenLibraryInternal(
     {
         TChar path[PAL_MAX_PATH_SIZE];
         path[0] = '\0';
-        Shlib_Format(path, self->providerDir, proventry->libraryName);
+        Shlib_Format(path, self->providerDir,
+                     proventry->interpreter ? "ScriptProvider" :
```
paulcallen: Wondering if this should be omiscriptprovider in case we have another provider that matches this.

Unix/provmgr/provmgr.c
```diff
@@ -2052,7 +2115,6 @@ MI_Result ProvMgr_Init(
     self->timeoutHandler.sock = INVALID_SOCK;
     self->timeoutHandler.data = self;
     self->timeoutHandler.callback = _TimeoutCallback;
-    self->timeoutHandler.handlerName = MI_T("PROVMGR_TIMEOUT_MANAGER");
```
paulcallen: You need to rebase, I added this and now it is gone

Unix/omireg/omireg.cpp
```diff
@@ -429,6 +473,43 @@ static MI_Module* LoadModule(const char* path)
     return module;
 }
 
+static MI_Module* LoadModuleFromScript(
+    const char* const interpreter,
+    const char* const startup,
+    const char* const moduleName)
+{
+    const char START[] = "Start";
+    const char PROVIDER[] = "libScriptProvider.so";
```
paulcallen: if we change the name, this would be changed too

Unix/provmgr/provmgr.c
EMumau: I did do a rebase.  Actually, I created a new branch so that I could cherry pick these changes.  I missed this change when I copied it over.

paulcallen:
@Microsoft/omi-devs - should get a review from someone else on the team as well.

Unix/provmgr/provmgr.c
```diff
@@ -190,7 +196,8 @@ static Library* MI_CALL _OpenLibraryInternal(
 
             if (!p->handle)
             {
-                trace_SharedLib_CannotOpenSecondTry(scs(proventry->libraryName), tcs(Shlib_Err()));
+                trace_SharedLib_CannotOpenSecondTry(
+                    scs(proventry->libraryName), tcs(Shlib_Err()));
```
JumpingYang001: looks this doesn't same as previous "proventry->interpreter ? "OMIScriptProvider" :  proventry->libraryName", suggest to new a TChar* var1=proventry->interpreter ? "OMIScriptProvider" :  proventry->libraryName; in the begin of function "static Library* MI_CALL _OpenLibraryInternal("

Unix/configure
```diff
@@ -2298,6 +2300,12 @@ then
     cp $rootrelative/etc/omigen.conf \$destdir/$sysconfdir/omigen.conf
 fi
 
+
+if test ! -f "\$destdir/$sysconfdir/omireg.conf"
```
jeffaco: This is an odd syntax. How about what we usually do, something like:

```
if [ ! -f "\$destdir/$sysconfdir/omireg.conf" ]; then
```

(and delete next line)

Unix/configure
```diff
@@ -2298,6 +2300,12 @@ then
...
+then 
+    cp $rootrelative/etc/omireg.conf \$destdir/$sysconfdir/omireg.conf
+fi
+
```
jeffaco: I'm not fully clear on why this is conditional, since at line 6383 you seem to be copying unconditionally. Can you explain why this is conditional please?

EMumau: All of the edits in configure are cut-n-paste copies of how omicli.conf, omigen.conf, and omiserver.conf are handled.  The behavior of the configuration is the same for every other conf file.

Unix/configure
EMumau: This is the same syntax used for omiserver.conf, omicli.conf, and omigen.conf.  It was a cut-n-paste.  This is how it is usually done.

FawcettJohnW: Both of these commands are in text that is output from this file, configure, to config.mak.  That file has two separate sections for the copy, one for dev that does and unconditional copy but uses -p to preserve times, while the other only adds if the old wasn't there.  These are config files, so the assumption is that users may well have changed them after the initial install and we don't mess with them on upgrade.

PR #335, opened 2017-06-22
Fix the problem with http encrypt and enum

yakman2020:
There was a problem with the server encrypt data, and one with client encrypt data. These have been fixed.


yakman2020:
@Microsoft/omi-devs  pls review

yakman2020:
Changes made as requested.

PR #336, opened 2017-06-22
nonRoot feature, with server split into server and engine.

palladia:
@Microsoft/omi-devs

### DIFF
--- a/Unix/build.mak
+++ b/Unix/build.mak
@@ -102,7 +102,7 @@ endif
 ifdef CONFIG_SINGLEIMAGE
   DIRECTORIES += image
 else
-  DIRECTORIES += server agent
+  DIRECTORIES += server agent engine
 endif
 
 ifeq ($(ENABLE_NATIVE_KITS),1)
@@ -274,7 +274,7 @@ DISTTAR=$(CONFIGUREDIR)/$(DIST).tar
 DISTTestProviders=samples/Providers
 
 distcommon:
-	@ rm -rf $(DISTTMPDIR)
+	@ sudo rm -rf $(DISTTMPDIR)
 	@ rm -f $(DISTTAR).gz
 	@ mkdir -p $(DISTTMPDIR)/$(DIST)
 	@ sudo cp -rp . $(DISTTMPDIR)/$(DIST)/
--- a/Unix/engine/GNUmakefile
+++ b/Unix/engine/GNUmakefile
@@ -7,11 +7,11 @@ include $(TOP)/config.mak
 ##
 
 ifdef HPUX_PARISC_HP
-  CXXPROGRAM = omiserver
-  SOURCES = server.c servermaincxx.cpp servercommon.c
+  CXXPROGRAM = omiengine
+  SOURCES = engine.c enginemaincxx.cpp ../server/servercommon.c
 else
-  CPROGRAM = omiserver
-  SOURCES = server.c servermain.c servercommon.c
+  CPROGRAM = omiengine
+  SOURCES = engine.c enginemain.c ../server/servercommon.c
 endif
 
 INCLUDES = $(TOP) $(TOP)/common
--- a/Unix/engine/engine.c
+++ b/Unix/engine/engine.c
@@ -1,0 +1,119 @@
+/*
+**==============================================================================
+**
+** Copyright (c) Microsoft Corporation. All rights reserved. See file LICENSE
+** for license information.
+**
+**==============================================================================
+*/
+
+#include <server/server.h>
+
+static Options s_opts;
+static ServerData s_data;
+
+int enginemain(int argc, const char* argv[])
+{
+    int pidfile = -1;
+    MI_Result result;
+
+    SetDefaults(&s_opts, &s_data, argv[0], OMI_ENGINE);
+
+    /* Get --destdir command-line option */
+    GetCommandLineDestDirOption(&argc, argv);
+
+    /* Extract configuration file options */
+    GetConfigFileOptions();
+
+    /* Extract command-line options a second time (to override) */
+    GetCommandLineOptions(&argc, argv);
+
+    /* Open the log file */
+    OpenLogFile();
+
+#if defined(CONFIG_POSIX)
+
+    /* ATTN: unit-test support; should be removed/ifdefed later */
+    if (s_opts.ignoreAuthentication)
+    {
+        IgnoreAuthCalls(1);
+    }
+
+    /* Watch for SIGTERM signals */
+    if (0 != SetSignalHandler(SIGTERM, HandleSIGTERM) ||
+        0 != SetSignalHandler(SIGHUP, HandleSIGHUP) ||
+        0 != SetSignalHandler(SIGUSR1, HandleSIGUSR1))
+    {
+        err(ZT("cannot set sighandler, errno %d"), errno);
+    }
+
+    /* Watch for SIGCHLD signals */
+    SetSignalHandler(SIGCHLD, HandleSIGCHLD);
+
+#endif
+
+    /* Change directory to 'rundir' */
+    if (Chdir(OMI_GetPath(ID_RUNDIR)) != 0)
+    {
+        err(ZT("failed to change directory to: %s"), scs(OMI_GetPath(ID_RUNDIR)));
+    }
+
+#if defined(CONFIG_POSIX)
+    /* Daemonize */
+    if (s_opts.daemonize && Process_Daemonize() != 0)
+    {
+        err(ZT("failed to daemonize engine process"));
+    }
+#endif
+
+    while (!s_data.terminated)
+    {
+        MI_Boolean r;
+
+        result = InitializeNetwork();
+        if (result != MI_RESULT_OK)
+        {
+            err(ZT("Failed to initialize network"));
+        }
+
+        result = WsmanProtocolListen();
+        if (result != MI_RESULT_OK)
+        {
+            err(ZT("Failed to initialize Wsman"));
+        }
+
+        // binary connection with server
+        result = BinaryProtocolListenSock(s_opts.socketpairPort, &s_data.mux[1], &s_data.protocol1, NULL, NULL);
+        if (result != MI_RESULT_OK)
+        {
+            err(ZT("Failed to initialize binary protocol for socket"));
+        }
+
+        r = SendSocketFileRequest(&s_data.protocol1->protocolSocket);
+        if (r == MI_FALSE)
+        {
+            err(ZT("failed to send socket file request"));
+        }
+
+        // Give it a little time for SocketFile info to come back from server
+        // Sleep_Milliseconds(50);
+        
+        // binary connection with client
+        const char *path = OMI_GetPath(ID_SOCKETFILE);
+        result = BinaryProtocolListenFile(path, &s_data.mux[0], &s_data.protocol0, NULL);
+        if (result != MI_RESULT_OK)
+        {
+            err(ZT("Failed to initialize binary protocol for socket file"));
+        }
+
+        result = RunProtocol();
+        if (result != MI_RESULT_OK)
+        {
+            err(ZT("Failed protocol loop"));
+        }
+    }
+
+    ServerCleanup(pidfile);
+
+    return 0;
+}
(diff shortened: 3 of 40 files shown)


Unix/protocol/protocol.c
```diff
@@ -440,6 +469,96 @@ static StrandFT _ProtocolSocket_FT = {
...
+
+static MI_Result _ProtocolSocketTrackerRemoveElement(Sock s)
+{
+    TrackerBucket b;
+    int r;
+
+    b.key = s;
+
+    Lock_Acquire(&s_trackerLock);
+    r = HashMap_Remove(&s_protocolSocketTracker, (HashBucket*)&b);
```
paulcallen: how does the bucket get deleted?

Unix/protocol/protocol.c
```diff
@@ -440,6 +469,96 @@ static StrandFT _ProtocolSocket_FT = {
...
+    b->key = s;
+    b->value = p;
+
+    Lock_Acquire(&s_trackerLock);
+    r = HashMap_Insert(&s_protocolSocketTracker, (HashBucket*)b);
+    Lock_Release(&s_trackerLock);
+
+    if (r != 0)
+    {
+        // already exists
```
paulcallen: can insert fail?
Also cant hash maps cause hash collisions. I assume exist does actually mean it exists, not just the hash that exists.

Unix/protocol/protocol.c
```diff
@@ -440,6 +469,96 @@ static StrandFT _ProtocolSocket_FT = {
...
+        trace_TrackerHashMapAlreadyExists(p, s);
+        MI_Result res = _ProtocolSocketTrackerRemoveElement(s);
+        if (res == MI_RESULT_OK)
+        {
+            Lock_Acquire(&s_trackerLock);
+            r = HashMap_Insert(&s_protocolSocketTracker, (HashBucket*)b);
```
paulcallen: can it fail?

Unix/protocol/protocol.c
```diff
@@ -490,8 +615,16 @@ static MI_Boolean _SendAuthRequest(
...
-    req->uid = geteuid();
-    req->gid = getegid();
+    if (uid != INVALID_ID && gid != INVALID_ID)
+    {
+        req->uid = uid;
+        req->gid = gid;
+    }
+    else
+    {
+        req->uid = geteuid();
```
paulcallen: which process are we in here? Why would we get invalid ids here? it does not seem to make sense to do this

Unix/protocol/protocol.c
palladia: HashMap_Remove() does it.

Unix/protocol/protocol.c
palladia: The only failure case is if r == 1, which means that the bucket exists.

Unix/protocol/protocol.c
palladia: It could.  And in that case, this AddElement function will return MI_RESULT_FAILED.

Unix/protocol/protocol.c
palladia: Only client would call _SendAuthRequest with invalid ids.

Unix/build.mak
```diff
@@ -102,7 +102,7 @@ endif
 ifdef CONFIG_SINGLEIMAGE
   DIRECTORIES += image
 else
-  DIRECTORIES += server agent
+  DIRECTORIES += server agent engine
```
jeffaco: So it looks like there's a new binary here, but it hasn't been put into the kit. This means that the RPM and DPKG kits that we build won't be valid, and will install a non-working version of OMI.

Note that PBUILD currently only builds --dev, not kits. Building kits are done on a restricted set of machines (just SLES 10 for Linux) due to universal binaries. You minimally need to make the `installbuilder` changes, build new kits on a SLES 10 system, and then test it to make sure that the version you install actually works.

I'm going to need this to be done prior to signing off this change.

Unix/engine/GNUmakefile
```diff
@@ -0,0 +1,39 @@
...
+##
+## Note: on HPUX_PARISC_HP, the server must be linked as C++, since we have
+## not found a way to dynamically load a C++ provider from a C program.
+##
+
+ifdef HPUX_PARISC_HP
+  CXXPROGRAM = omiengine
+  SOURCES = engine.c enginemaincxx.cpp ../server/servercommon.c
+else
+  CPROGRAM = omiengine
```
jeffaco: So binary file `omiengine` should now exist in `/opt/omi/bin`?

Unix/engine/engine.c
```diff
@@ -0,0 +1,119 @@
...
+    /* Extract configuration file options */
+    GetConfigFileOptions();
+
+    /* Extract command-line options a second time (to override) */
+    GetCommandLineOptions(&argc, argv);
+
+    /* Open the log file */
+    OpenLogFile();
+
+#if defined(CONFIG_POSIX)
```
jeffaco: Hmm. When would we NOT be using POSIX? What's the purpose if this `#if`?

Unix/engine/engine.c
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        r = SendSocketFileRequest(&s_data.protocol1->protocolSocket);
+        if (r == MI_FALSE)
+        {
+            err(ZT("failed to send socket file request"));
+        }
+
+        // Give it a little time for SocketFile info to come back from server
+        // Sleep_Milliseconds(50);
```
jeffaco: This is here why? Do we need extra time or not? What if the system is heavily loaded?

Unix/engine/engine.c
```diff
@@ -0,0 +1,119 @@
...
+**
+**==============================================================================
+*/
+
+#include <server/server.h>
+
+static Options s_opts;
+static ServerData s_data;
+
+int enginemain(int argc, const char* argv[])
```
jeffaco: No common header file, and this wasn't extern "C"'d.  What happens if `Unix/engine/enginemaincxx.cpp` is used (I guess on HP-UX, for whatever reason)?

I'm guessing you didn't test this on that platform? Does it even pass `PBUILD`?

Unix/etc/omiserver.conf
```diff
@@ -58,3 +58,5 @@ NoSSLCompression=false
 #NtlmCredsFile=PATH
 # For example
 #NtlmCredsFile=/etc/opt/omi/.creds/ntlm
+nonroot=false
+service=test1
```
jeffaco: I guess this is a "fake" omiserver.conf file (not the one we ship to customers). But still, do we really want lines 61/62 here? Shouldn't unit tests set this up if that's what they want?

Unix/protocol/protocol.c
```diff
@@ -8,6 +8,9 @@
 */
 
 #include <assert.h>
+#include <sys/types.h>
```
jeffaco: I don't fully understand this code ... @yakman2020 can you please review this too? I'd like to see Bruce review this prior to checkin please.

Unix/provreg/provreg.c
```diff
@@ -511,7 +511,7 @@ static int _AddEntry(
 #endif /* defined(CONFIG_ENABLE_PREEXEC) */
 
     /* ProvRegEntry.hosting*/
-    e->hosting = PROV_HOSTING_INPROC;
+    e->hosting = PROV_HOSTING_REQUESTOR;
```
jeffaco: So is this changing the default hosting type to `@requestor@`, assuming that the `HOSTING=` line isn't specified?

Where is that parsed from the .reg file? Is `inproc` disabled now? Or will it be a separate commit?

If a separate commit, what's the purpose of the changes here?

Unix/provreg/provreg.c
```diff
@@ -528,7 +528,7 @@ static int _AddEntry(
          */
         if (strcmp(hosting, PROV_REG_HOSTING_INPROC) == 0)
         {
-            e->hosting = PROV_HOSTING_INPROC;
+            e->hosting = PROV_HOSTING_REQUESTOR;
```
jeffaco: Yuck. Why not just make INPROC be invalid, and return an error? I see no reason why asking, explicitly, for INPROC should not be an error and would do something completely different ...

Please explain your rationale, thanks.

Unix/tests/wsman/utils.cpp
```diff
@@ -308,6 +309,9 @@ int StartServerAndConnect(
 
     UT_ASSERT( MI_RESULT_OK == result );
 
+    // A new socket connection is created, give it a little time
+    ut::sleep_ms(200);
```
jeffaco: I hate this kind of thing. What if the test system is slower than you expect, for whatever reason? Why not wait for the socket for some "reasonable" period of time, like this:

```
int tries = 0;
while ( true )
{
    if (socket_exists)
    {
        break;
    }

    if (++tries > 10);
    {
        emit error and abort;
    }

    ut:sleep_ms(100);
}
```

This is much more resilient, even if the machine is slower than expected for whatever reason.

jeffaco: This sort of thing is true for other hard-coded sleep statements above this too ...

Unix/etc/omiserver.conf
jeffaco: Should defaults be added to the "real" omiserver.conf file shipping to customers on a kit install?

Unix/build.mak
palladia: Fixed installbuilder/datafiles/Base_OMI.data

Unix/engine/GNUmakefile
palladia: yes

Unix/engine/engine.c
palladia: pbuild tested on HP

Unix/engine/engine.c
palladia: leftover from days of supporting Windows.  If we want to remove Windows support, we can go in and clean them up.

Unix/provreg/provreg.c
```diff
@@ -615,7 +616,7 @@ static int _AddEntryForExtraClass(
          */
         if (strcmp(regClass->hosting, PROV_REG_HOSTING_INPROC) == 0)
         {
-            hosting = PROV_HOSTING_INPROC;
+            hosting = PROV_HOSTING_REQUESTOR;
```
paulcallen: Overriding inproc silently does not sound right. You definitely need a very big error telling which provider asked for inproc, but it was made out of processes and that the provider probably needs to be updated.

Unix/http/httpauth.c
```diff
@@ -1639,19 +1711,52 @@ MI_Boolean IsClientAuthorized(_In_ Http_SR_SocketData * handler)
...
+                trace_HTTP_UserAuthFailed("user not authenticated");
+                handler->authFailed = TRUE;
+                
+                _SendAuthResponse(handler, auth_response, response_len);
+                return FALSE;
+            }
+        }
+        else
+        {
+            if (0 != _AskServerToAuthenticateUser(headers->username,
```
paulcallen: OK, I think we need to make this asynchronous and use the binary protocol properly, sending a proper PamAuthResponse from the server and break this code apart so the code after this point happens as part of the PanAuthResponse.
We need to do this because we are on the main Protocol_Run thread and we are blocking that thread from processing other requests during this time

Unix/http/httpauth.c
```diff
@@ -1639,19 +1711,52 @@ MI_Boolean IsClientAuthorized(_In_ Http_SR_SocketData * handler)
...
+            if (0 != AuthenticateUser(headers->username, headers->password))
+            {
+                handler->httpErrorCode = HTTP_ERROR_CODE_UNAUTHORIZED;
+                auth_response = (char *)RESPONSE_HEADER_UNAUTH_FMT;
+                response_len  = RESPONSE_HEADER_UNAUTH_FMT_LEN;
+
+                trace_HTTP_UserAuthFailed("user not authenticated");
+                handler->authFailed = TRUE;
+                
+                _SendAuthResponse(handler, auth_response, response_len);
```
palladia: Not sure which failure path you are referring to.  These blocks of code deal with failure cases.  If you mean _SendAuthResponse...it's a void function, but does log if it fails.

PR #340, opened 2017-06-29
defunct: Adding features to OMI to support OMI-Script-Provider.

EMumau:


paulcallen:
looks like you need to rebase, and you will need to squash your commit list into a single checking commit.

paulcallen:
also, if they python stuff is not supported on any specific platform you may want to #ifdef the changes in the provider manager, and fail the registration on those platforms.

JumpingYang001:
We should hold on this PR to wait non-root feature release then we can merge it and start Python Provider building and testing, thanks.

JumpingYang001:
I will share here once non-root release then we can check it that time.
BTW, looks like omireg doesn't support c/c++ providers in your code:
```
root@py64-cent7-01  # /opt/omi/bin/omireg --hosting '@requestor@' -n root/cimv2 /root/Build-omi-script-provider/omi/Unix/output/lib/libScriptProvider.so
Created /opt/omi/lib/libScriptProvider.so
/opt/omi/bin/omireg: failed to find symbol 'MI_Main' in library '/root/Build-omi-script-provider/omi/Unix/output/lib/libScriptProvider.so'

```

PR #343, opened 2017-06-30
Fix seg fault on some platforms

palladia:
It was observed on pbuild of osdev64-cent5-01 and osd64-sls10-01 that seg fault happens during omiserver shutdown.

Investigations show that during shutdown, when Selector_RemoveAllHandlers() is called, a sequence of _ProtocolSocket_* functions are invoked:

_ProtocolSocket_Cleanup()
_ProtocolSocket_Close()
_ProtocolSocket_Finish()
_ProtocolSocket_Release()

Since both _ProtocolSocket_Finish() and _ProtocolSocket_Release() frees the allocated memory, one shouldn't call both.  The fix for now is to remove the _ProtocolSocket_Release() line.


palladia:
@Microsoft/ostc-devs

PR #349, opened 2017-07-10
Disable replay cache temporarily

JumpingYang001:
@Microsoft/omi-devs , I have verified on Ubuntu 16.04 and the leak open too many files issue disappear, thanks. Could you help to review this PR? thanks very much!


yakman2020:
1.IF we were going to set an environment variable, we would do it somewhere it would happen once. This is happening quite often and would be very inefficient.

2.But setting an environment variable inside the code at all isn't a good idea. There is a security cost to turning off the replay cache: krb5 would be unable to detect and defend from replay attacks. So this needs to be decided by the administrator.

3. The right place to debug this looks like in krb5 itself where it is losing the handles. I'm looking to see if the krb5 people have already done this, and if not we need to work with the platform owners to get it fixed.

PR #362, opened 2017-07-31
Make PAM authentication asynchronous

palladia:


Unix/http/httpauth.c
```diff
@@ -1605,83 +1612,70 @@ void Deauthorize(_In_ Http_SR_SocketData * handler)
...
-    GenerateRandomString(rString, sizeof(rString)); 
+    Http_SR_SocketData *handler = (Http_SR_SocketData*)(uintptr_t)msg->handle;
+    HttpHeaders *headers = &handler->recvHeaders;
 
-    Strlcpy(file, OMI_GetPath(ID_SYSCONFDIR), PAL_MAX_PATH_SIZE);
-    Strlcat(file, "/sockets/", PAL_MAX_PATH_SIZE);
-    Strlcat(file, rString, PAL_MAX_PATH_SIZE);
+    // Resume selector processing of this socket
+    handler->handler.mask |= SELECTOR_READ;
+    Selector_Wakeup(handler->http->selector, MI_TRUE);
```
paulcallen: need to do this AFTER you set the auth members on the handler otherwise the pam authorization may happen again

Unix/protocol/protocol.c
```diff
@@ -1310,18 +1314,18 @@ static MI_Boolean _ProcessCreateAgentMsg(
...
     const char *user,
     const char *passwd,
-    const char *file
+    MI_Uint64 handle
     )
 {
-    PamCheckUserMsg *req = NULL;
+    PamCheckUserReq *req = NULL;
     MI_Boolean retVal = MI_TRUE;
+    ProtocolSocket *h = s_permanentSocket;
```
paulcallen: Can you rename this stack variable please! 'h' confused me greatly. Using the global variable directly would have been more obvious

Unix/http/httpauth.c
```diff
@@ -1605,83 +1612,69 @@ void Deauthorize(_In_ Http_SR_SocketData * handler)
...
+    memset(&handler->recvHeaders, 0, sizeof(handler->recvHeaders));
+    handler->recvingState = RECV_STATE_HEADER;
 
-    /* Add user to cache if auth was ok */
-    if (0 == r)
-        CredCache_PutUser(user, password);
-
-    return r;
+    // Resume selector processing of this socket
+    handler->handler.mask |= SELECTOR_READ;
```
paulcallen: move right before Process_Authorized_Message. Not set for the error cases as they set the write state to send the error. Also, Process_Authorized_Message will ultimately set the write state to send the response back

Issue #370, opened 2017-08-07
"service_control restart" doesn't start the process on Aix


JumpingYang001:
Run `/opt/omi/bin/service_control restart` on ostst16-aix71-04 when omiserver process is active.
You will find the process doesn't start, that's because Aix is a slow operating system or machine, when it is enough slow, when 'stop_omi' function completed but the omi pid does still exist, then you will get 0 for $? here: https://github.com/Microsoft/omi/blob/master/Unix/installbuilder/service_scripts/service_control.unix#L100 .

I have tried to add `sleep 1` after here: https://github.com/Microsoft/omi/blob/master/Unix/installbuilder/service_scripts/service_control.unix#L145, then it works fine.

Repro command and result:
```
 /root
 root@ostst16-aix71-04 # /opt/omi/bin/service_control restart
0513-044 The omid Subsystem was requested to stop.

 /root
 root@ostst16-aix71-04 # ps -ef|grep omi

 /root
 root@ostst16-aix71-04 #
```

JumpingYang001:
It is fixed with #371 , close it, thanks.

PR #371, opened 2017-08-07
Fix service_control restart on Aix

JumpingYang001:
Fixed issue #370 , @Microsoft/omi-devs , could you help to review this PR? thanks very much!

### DIFF
--- a/Unix/installbuilder/service_scripts/service_control.unix
+++ b/Unix/installbuilder/service_scripts/service_control.unix
@@ -90,6 +90,18 @@ solaris_wait_service_stop()
     done
 }
 
+wait_service_stop()
+{
+    COUNT=0
+    while [ $COUNT -lt 15 ]; do
+	echo "Waiting for service stop ..."
+	sleep 1
+	is_omi_running
+	[ $? -ne 0 ] && break
+	COUNT=`expr $COUNT + 1`
+    done
+}
+
 #
 # Normal usage functions (used by everything except service control scripts)
 #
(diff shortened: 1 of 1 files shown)


Unix/installbuilder/service_scripts/service_control.unix
```diff
@@ -143,6 +143,7 @@ restart_omi()
 
     if [ "$TYPE" = "A" -o "$TYPE" = "H" -o "$TYPE" = "S" ]; then
 	stop_omi
+	sleep 1
```
jeffaco: Blech, I hate this kind of thing. It's obviously a timing problem, which is why we haven't seen it before. What if the system is slow enough where 1 second isn't long enough?

Please actively checking if OMI is down and then wait only if OMI is still running. We've done this sort of thing in other areas of code, please use a similar method here. Thanks.

JumpingYang001:
@jeffaco , I have fixed based on your comments, could you help to review it again? thanks very much!

Unix/installbuilder/service_scripts/service_control.unix
```diff
@@ -143,6 +155,7 @@ restart_omi()
 
     if [ "$TYPE" = "A" -o "$TYPE" = "H" -o "$TYPE" = "S" ]; then
 	stop_omi
+	wait_service_stop
```
jeffaco: Since this is only called from one place, you didn't need to make it a subroutine. But this is a style choice, I have no problem with the change. Signing off.

PR #382, opened 2017-08-17
Fix memory leak in protocol.c and wsman.c

palladia:
Fixes following:

1.  Leak with ProtocolSocket where refCount wasn't decremented
2.  Leak with ProtocolSocket where strand->otherClosedThis wasn't set
3.  Leak with wsman messages where options weren't deleted
4.  Changed trace print message for authorization failures to "Error" category so that they can be seen in log files more easily.

### DIFF
--- a/Unix/protocol/protocol.c
+++ b/Unix/protocol/protocol.c
@@ -772,8 +772,11 @@ static MI_Boolean _ProcessAuthMessageWaitingConnectRequestFileData(
 */
 static MI_Boolean _ProcessAuthMessageWaitingConnectRequest(
     ProtocolSocket* handler,
-    BinProtocolNotification* binMsg)
+    BinProtocolNotification* binMsg,
+    MI_Boolean *keepConnection)
 {
+    *keepConnection = MI_FALSE;
+
     /* un-expected message */
     if (BinNotificationConnectRequest != binMsg->type)
         return MI_FALSE;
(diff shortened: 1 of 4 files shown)


Unix/wsman/wsman.c
```diff
@@ -1616,6 +1616,8 @@ static void _ProcessEnumerateRequest(
     if (!msg->className || !msg->nameSpace)
     {
         _CD_ProcessEnumFailed( selfCD, enumContext );
+        __MI_Instance_Destruct(msg->base.options);
+        msg->base.options = NULL;
```
paulcallen: Shouldn't EnuemerateInstance_Req_Release clean out the options? They are owned by that object.

Unix/protocol/protocol.c
```diff
@@ -912,13 +920,31 @@ static MI_Boolean _ProcessAuthMessage(
...
     if (PRT_AUTH_WAIT_CONNECTION_REQUEST == handler->clientAuthState)
     {
-        return _ProcessAuthMessageWaitingConnectRequest(handler, binMsg);
+        MI_Boolean keepConnection;
+        MI_Boolean r;
+
+        r = _ProcessAuthMessageWaitingConnectRequest(handler, binMsg, &keepConnection);
+        if (r && !keepConnection)
+        {
+            handler->strand.info.otherClosedThis = MI_TRUE;
```
paulcallen: unfortunately you should not exit these directly. This implies you need to call Strand_Close somewhere

Unix/wsman/wsman.c
```diff
@@ -1768,6 +1780,8 @@ static void _ProcessAssociatorsRequest(
 
     _OpenRightEnum(selfCD,enumContext,&msg->base,updateTimer);
 
+    __MI_Instance_Destruct(msg->base.options);
+    msg->base.options = NULL;
```
paulcallen: same goes for all the other operations you fixed.

PR #385, opened 2017-08-29
Fix memory leak in IsClientAuthorized()

palladia:


paulcallen:
@yakman2020 can you review this

yakman2020:
Did

PR #401, opened 2017-09-08
Update Kerberos doc for keytab

JumpingYang001:
@Microsoft/omi-devs , I have updated Kerberos doc, could you help to review this PR? thanks very much!

@yakman2020, as we discussed, it might be better to add these commands in doc, could you help to review my PR? thanks very much for help!


JumpingYang001:
thanks for approving! will squash tomorrow.

PR #409, opened 2017-10-05
Preexec must be executed as root

paulcallen:
- proxy preexec requests to the omiserver process from the omiengine process so it can be executed as root
- pass logging file handle to omiengine so it has permission to log

@Microsoft/omi-devs

### DIFF
--- a/Unix/base/messages.h
+++ b/Unix/base/messages.h
@@ -81,6 +81,11 @@ typedef enum _MessageTag
     VerifySocketConnTag = 36,
     PamCheckUserReqTag = 37,
     PamCheckUserRespTag = 38
+#if defined(CONFIG_ENABLE_PREEXEC)
+    ,
+    ExecPreexecReqTag   = 39,
+    ExecPreexecRespTag  = 40
+#endif
 }
 MessageTag;
 
@@ -1246,6 +1251,14 @@ MI_Boolean Message_IsInternalMessage( _In_ Message* msg )
     if (PamCheckUserRespTag == msg->tag)
         return MI_TRUE;
 
+#if defined(CONFIG_ENABLE_PREEXEC)
+    if (ExecPreexecReqTag == msg->tag)
+        return MI_TRUE;
+
+    if (ExecPreexecRespTag == msg->tag)
+        return MI_TRUE;
+#endif
+
     return MI_FALSE;
 }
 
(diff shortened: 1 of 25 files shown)


Unix/base/messages.h
```diff
@@ -1814,6 +1827,98 @@ MI_INLINE void __PamCheckUserResp_Release(
...
+    __ExecPreexecReq_Release(self, CALLSITE)
+
+MI_INLINE void __ExecPreexecReq_Release(
+    ExecPreexecReq* self,
+    CallSite cs)
+{
+    __Message_Release(&self->base, cs);
+}
+
+void ExecPreexecReq_Print(const ExecPreexecReq* msg, FILE* os);
```
palladia: Should define this in base/messageprint.c, and add a "case" in MessagePrint.

Unix/base/messages.h
```diff
@@ -1814,6 +1827,98 @@ MI_INLINE void __PamCheckUserResp_Release(
...
+    __ExecPreexecResp_Release(self, CALLSITE)
+
+MI_INLINE void __ExecPreexecResp_Release(
+    ExecPreexecResp* self,
+    CallSite cs)
+{
+    __Message_Release(&self->base, cs);
+}
+
+void ExecPreexecResp_Print(const ExecPreexecResp* msg, FILE* os);
```
palladia: Add this in base/messagesprint.c also.

Unix/configure
```diff
@@ -543,6 +543,7 @@ fi
 outputdir=$configuredir/$outputdirname
 
 if [ "$dev" = "1" ]; then
+  enable_preexec=1
```
palladia: Why make this feature optional?  And why only for dev builds?

Unix/tests/cli/test_cli.cpp
```diff
@@ -74,6 +75,100 @@ namespace
     bool runKrbTests       = false;
 }
 
+
+
+// Post Mortem log handling. Maybe should be in nits base.
```
palladia: I am confused.  I thought we'd dump the log for the whole travis run, not just for cli tests.  This would help for preexec, but not if failure occurs in wsman, protocol, etc.?

Unix/configure
yakman2020: preexec is enabled for --microsoft also. We only add it to dev

Unix/tests/cli/test_cli.cpp
yakman2020: When the code runs correctly we can reshuffle a bit. (for the next release)

PR #411, opened 2017-10-11
Modified shutdown/restart logic

palladia:
If engine dies, kill server.
If server dies, kill engine and agents.
Auto-restart logic removed.

### DIFF
--- a/Unix/agent/agent.c
+++ b/Unix/agent/agent.c
@@ -326,8 +326,11 @@ int agent_main(int argc, const char* argv[])
     MI_Result r;
     Sock fd;
     int logfd;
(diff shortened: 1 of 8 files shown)


palladia:
@Microsoft/omi-devs

jeffaco:
Did you, by chance, build a kit and install it on "modern" systems (systems with `systemd`) to insure that the server is restarted appropriately? There are some assumptions here on how `systemd` behaves, we should validate them and insure proper behavior (or look at changing the `systemd` unit file appropriately).

Please do that prior to committing this, thanks.

palladia:
tested a kit on a redhat 7 system

PR #417, opened 2017-10-19
Fixes crontab and gss lib link issues

yakman2020:
Checked on sls and ubuntu.

### DIFF
--- a/Unix/scripts/config_keytab_update.sh
+++ b/Unix/scripts/config_keytab_update.sh
@@ -54,7 +54,9 @@ unconfigure()
 
     crontab -l > $tmpfile 2> /dev/null || true
 
-    egrep -v "$omikeytab" $tmpfile | crontab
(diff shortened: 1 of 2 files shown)


yakman2020:
@Microsoft/omi-devs

Unix/scripts/config_keytab_update.sh
```diff
@@ -54,7 +54,9 @@ unconfigure()
 
     crontab -l > $tmpfile 2> /dev/null || true
 
-    egrep -v "$omikeytab" $tmpfile | crontab
+    tmpfile2=$(mktemp)
+    egrep -v "$omikeytab" $tmpfile >$tmpfile2
```
jeffaco: Change this to `grep`, you're not using regular expressions here.

It's really dumb, on the Sles-10 platform, that we get comments in the file. But I don't see a way around this.

PR #423, opened 2017-10-30
Fix problem with non-dying agents

palladia:
Need to distinguish normal timeout from idle timeout.


JumpingYang001:
I am not sure if the error in the else need break or not, please consider it, thanks.

JumpingYang001:
ok, got it, thanks.

PR #428, opened 2017-11-02
Fixes unit test issue caused by increasing max header size

yakman2020:
@Microsoft/omi-devs 

pbuild passed

### DIFF
--- a/Unix/tests/http/test_http.cpp
+++ b/Unix/tests/http/test_http.cpp
@@ -15,6 +15,7 @@
 #include <base/result.h>
 #include <base/Strand.h>
 #include <base/log.h>
+#include <http/http_private.h>
 
 static MI_Uint16 PORT = ut::getUnittestPortNumber() + 10;
 
@@ -784,7 +785,7 @@ NitsTestWithSetup(TestHttp_HeaderTooBig, TestHttpSetup)
         "Content-Length: 5\r\n"
         "Authorization: auth\r\n";
 
-    data += string("hugeAttribute: ") + string(4096, 'x') + string("\r\n");
+    data += string("hugeAttribute: ") + string(MAX_HEADER_SIZE, 'x') + string("\r\n");
     data += "\r\n"
         "Hello";
 


Unix/tests/http/test_http.cpp
```diff
@@ -22,6 +22,8 @@ static MI_Uint16 PORT = ut::getUnittestPortNumber() + 10;
 #define TEST_USERNAME "test"
 #define TEST_PASSWORD "password"
 
+static const MI_Uint32 MAX_HEADER_SIZE = 65536;
```
jeffaco: A magic number. I'd much prefer that you set a constant or #define in the source file, then use that here to size the string. That way, if header size is changed again, the unit test fixes itself automatically.

I think that's the "right" fix. But minimally I'd like a comment explaining where 65536 comes from, and what to do if there are problems. But resolving a const int or #define is much nicer to me.

yakman2020: Did you not notice that *is* a constant?

I replaced a magic number with a named constant.  In C++ #define is more or less deprecated for constant values because it produces no symbol.  I put it at the top of the test source with a name so it can be changed as needed. 

I don't see how that differs from what you are saying:

".But resolving a const int or #define is much nicer to me."
vs
static const MI_Uint32 MAX_HEADER_SIZE = 65536;

Note that MI_Uint32 is int (ok, unsigned int)

jeffaco: You're focusing on the wrong part. I'm not concerned with the left side of the `=`, I'm concerned of the right hand of the `=` sign.

`65536` is a magic number. Where does it come from? Can it be derived from the implementation itself?

yakman2020:
removed the constant and included http_private.h, the authoratative location of the define.

Unix/tests/http/test_http.cpp
```diff
@@ -784,7 +785,7 @@ NitsTestWithSetup(TestHttp_HeaderTooBig, TestHttpSetup)
         "Content-Length: 5\r\n"
         "Authorization: auth\r\n";
 
-    data += string("hugeAttribute: ") + string(4096, 'x') + string("\r\n");
+    data += string("hugeAttribute: ") + string( MAX_HEADER_SIZE, 'x') + string("\r\n");
```
jeffaco: Nit: remove space before `MAX_HEADER_SIZE`.

yakman2020: done

PR #430, opened 2017-11-06
Adding the MI_Module_Self parameter to the start function signature f…

EMumau:
Adding the MI_Module_Self parameter to the start function signature for the Script Provider.

This is the same change that was done to provmgr/provmgr.c in #427


JumpingYang001:
I have rebased from master, will merge it later, thanks.

PR #441, opened 2017-11-13
Resolve problems with psrp

yakman2020:
Resolves issues 437, 438, 439 

@Microsoft/omi-devs 
@dantraMSFT


paulcallen:
Client logging was broken because the getting of the path was messed up. That fixed that.
The +2 error was because did not account for the extra <cr><lf> that terminates the http header
The writeclientheaders should be able to be called multiple times, we failed incorrectly before, and now we return _continue if we are not still writing headers.

PR #446, opened 2017-11-17
Minor fix (proposal)

kant:


JumpingYang001:
I think the previous document is fine and just close this PR, if someone disagree, then he/she can add comments here.
Anyway, appreciate that you new a PR here, thank you!

PR #453, opened 2017-11-21
Fix NTLM ERROR to WARNING

### LINKED ISSUE #449: OMI Logging: OMI logs ERROR message (NTLM creds) in absolutely normal case
When starting up OMI v1.4.1-0 on a machine that is not domain joined (the normal case for many of our customers), I see an error in the OMI log file:

```
2017/11/17 15:00:44 [125770,125770] ERROR: null(0): EventId=20135 Priority=ERROR NTLM Credentials file does not exist or invalid permissions: null
```

OMI should not log error messages in normal cases. This causes needless OMI log growth and customer confusion.

JumpingYang001:
@jeffaco  , I have added a new PR here to fix #449 , could you help to review it? thank you!

### DIFF
--- a/Unix/base/oi_traces.h
+++ b/Unix/base/oi_traces.h
@@ -345,8 +345,6 @@ OI_EVENT("Reg file %s is corrupted. It will be skipped by the server")
 void trace_ProvReg_SkipRegFile(const char * filePath);
 OI_EVENT("LIBRARY tag is missing in reg file: %s")
 void trace_RegFile_MissingLibraryTag(const char * filePath);
-OI_EVENT("NTLM Credentials file does not exist or invalid permissions: %s")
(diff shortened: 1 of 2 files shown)


JumpingYang001:
@Microsoft/omi-devs

JumpingYang001:
Got it, thanks.

JumpingYang001:
@jeffaco , can I add a check with [getdomainname()](http://man7.org/linux/man-pages/man2/setdomainname.2.html) to determine if it is in the domain? or do you have other better option for checking it is joined the domain?
Checking the condition here: https://github.com/Microsoft/omi/blob/648f5c952849bf543050a966fcf23d2c5e5ec86b/Unix/server/server.c#L584

The error show only when omi start.

JumpingYang001:
I will new a PR if I know how to fix it, so now just close this PR, thanks.

PR #459, opened 2017-11-29
Enhancements to allow field dump of allocated memory

palladia:


palladia:
@Microsoft/omi-devs

Unix/base/batch.c
```diff
@@ -15,7 +15,12 @@
 #define BATCH_PAGE_SIZE ((size_t)(BLOCK_SIZE * 2))
 #define BLOCK_SIZE ((size_t)512)
 
-void* Batch_Get(
+void* __Batch_Get(
+#if defined(CONFIG_ENABLE_DEBUG) && defined(CONFIG_OS_LINUX)
```
jeffaco: Do you want to restrict this to LINUX? Does the memory dump code work on other platforms? If so, wouldn't we want this layered on top of that?

Unix/base/batch.c
```diff
@@ -35,7 +40,11 @@ void* Batch_Get(
     /* Allocate a new page and carve new block from that page */
     if (size8 <= BLOCK_SIZE && self->numPages < self->maxPages)
     {
+#if defined(CONFIG_ENABLE_DEBUG) && defined(CONFIG_OS_LINUX)
+        Page* page = __PAL_Malloc(file, line, func, BATCH_PAGE_SIZE);
```
jeffaco: Feel like file, line, and func should be passed after BATCH_PAGE_SIZE. BATCH_PAGE_SIZE is mandatory, the others get passed based on mode. Ditto for all below.

JumpingYang001: I don't find where we define __PAL_Malloc?

Unix/base/batch.h
```diff
@@ -113,68 +113,173 @@ void Batch_InitFromBuffer(
...
+    size_t line,
+    const char* func,
+    Batch* self,
+    const char* str)
+{
+    return __Batch_Strdup(file, line, func, self, str);
+}
+#endif
+
+#if defined(CONFIG_ENABLE_WCHAR)
```
jeffaco: Is WCHAR still a supported build?

Unix/pal/alloc.c
```diff
@@ -119,16 +120,12 @@ void PAL_DumpAllocList()
     if (!_list)
         return;
 
-    printf("\nWARNING: one or more blocks still allocated!\n");
+    trace_DumpAllocList_Warning();
```
jeffaco: What, exactly, does this do?

Unix/base/batch.c
JumpingYang001: ho, found it in pal allocate.c

Unix/base/oiomi.h
```diff
@@ -2471,6 +2471,18 @@ FILE_EVENT3(40038, trace_Product_Version_Impl, LOG_INFO, PAL_T("Starting %s: ver
...
+#define trace_DumpAllocList_Warning() trace_DumpAllocList_Warning_Impl(0, 0)
+#endif
+FILE_EVENT0(40040, trace_DumpAllocList_Warning_Impl, LOG_INFO, PAL_T("WARNING: one or more blocks still allocated!"))
+#if defined(CONFIG_ENABLE_DEBUG)
+#define trace_DumpAllocList_Block(a0, a1, a2, a3, a4, a5) trace_DumpAllocList_Block_Impl(__FILE__, __LINE__, scs(a0), a1, a2, a3, a4, a5)
+#else
+#define trace_DumpAllocList_Block(a0, a1, a2, a3, a4, a5) trace_DumpAllocList_Block_Impl(0, 0, scs(a0), a1, a2, a3, a4, a5)
+#endif
+FILE_EVENT6(40041, trace_DumpAllocList_Block_Impl, LOG_INFO, PAL_T("BLOCK: %s(%u): ptr=%p: magic=%08X id=%u size=%u"), const char*, unsigned int, const void*, unsigned int, unsigned int, unsigned int)
+#if defined(CONFIG_ENABLE_DEBUG)
```
jeffaco: I don't understand why you're adding these. Well, I see why you're adding them, but you wouldn't need to add these if you didn't make this Linux-specific. Why did you make this Linux-specific?

Unix/base/batch.c
palladia: If you look in pal/palcommon.h, you see that PAL_DumpAllocList() is defined only if USE_ALLOCATOR is defined, and USE_ALLOCATOR is defined if CONFIG_ENABLE_DEBUG and CONFIG_OS_LINUX are both defined.

Unix/base/batch.c
palladia: I'm following same convention as some existing functions.  See declaration of __PAL_Malloc and __PAL_Calloc in pal/palcommon.h.   If you feel strongly about this, I can move code around.

Unix/base/batch.h
palladia: We still have a lot of code based on CONFIG_ENABLE_WCHAR.  If you don't think we'll support it, I can remove it.

Unix/base/oiomi.h
palladia: These were added because the DumpAllocList() function was only dumping to standard output.  I talked to Paul about this, and decided that it's better to dump this information into the log file.

palladia: The disadvantage of this is that we'll have to enable loglevel=INFO or greater to see the dump.

Unix/pal/alloc.c
palladia: Not sure what you're asking here.  It's printing the same thing, but to the log file.

Unix/base/batch.h
```diff
@@ -113,68 +113,173 @@ void Batch_InitFromBuffer(
...
 _Check_return_
-ZChar* Batch_Tcsdup(
+char* __Batch_Strdup(
+    const char* file,
+    size_t line,
+    const char* func,
+    Batch* self,
+    const char* str);
+
+#if defined(CONFIG_ENABLE_WCHAR)
```
jeffaco: Definitely get rid of this stuff, not needed.

Unix/tools/gdb_allocdump
```diff
@@ -0,0 +1,3 @@
+call PAL_DumpAllocList()
+detach
+q
```
jeffaco: Spell this out as "quit".

Unix/base/batch.h
```diff
@@ -113,68 +113,173 @@ void Batch_InitFromBuffer(
...
-Batch* Batch_New(
-    size_t maxPages);
-
 void Batch_Delete(
     Batch* self);
 
-void* Batch_Get(
+#if defined(CONFIG_ENABLE_DEBUG) && defined(CONFIG_OS_LINUX)
+
+#define BATCH_CALLSITE __FILE__, __LINE__, __FUNCTION__
```
paulcallen: if you defined this with a trailing comma, you would not need to duplicate the Batch_* functions from debug and not

Unix/pal/alloc.c
```diff
@@ -119,16 +120,12 @@ void PAL_DumpAllocList()
     if (!_list)
         return;
 
-    printf("\nWARNING: one or more blocks still allocated!\n");
```
paulcallen: I would be tempted to leave the printf as well.

palladia: This would duplicate the output if omiserver is run with -l.

Unix/base/batch.h
palladia: Fixed.

palladia:
Script to get memory dump of all OMI processes will be done in a separate PR.

palladia:
@paulcallen

Unix/base/batch.h
```diff
@@ -113,68 +113,138 @@ void Batch_InitFromBuffer(
...
+#define Batch_Get(self, size) __Batch_Get(BATCH_CALLSITE self, size)
+#define Batch_GetClear(self, size) __Batch_GetClear(BATCH_CALLSITE self, size)
+#define Batch_Tcsdup(self, str) __Batch_Tcsdup(BATCH_CALLSITE self, str)
+#define Batch_Strdup(self, str) __Batch_Strdup(BATCH_CALLSITE self, str)
+#define Batch_StrTcsdup(self, str) __Batch_StrTcsdup(BATCH_CALLSITE self, str)
+#define Batch_ZStrdup(self, str) __Batch_ZStrdup(BATCH_CALLSITE self, str)
+
+#if defined(CONFIG_ENABLE_DEBUG) && defined(CONFIG_OS_LINUX)
+
+#define BATCH_CALLSITE __FILE__, __LINE__, __FUNCTION__,
```
paulcallen: if you had another define for lets say:
#define BATCH_CALLSITE_DEF const char * file, size_t line, const char *func,
and
#define BATCH_CALLSITE_PASSTHROUGH file, line, func, 
then I _think_ it would be more or less the same code throughout except the macros for debug get filled out and are empty when not.
A similar thing was done for strand debug code throughout.

PR #465, opened 2017-12-05
Add loglevel property in Configuring OMI Server

JumpingYang001:
@Microsoft/omi-devs , I have added loglevel property in Configuring OMI Server and update python provider status: "Remove '(Not ship yet)'", could you help to review this PR? thanks very much!


JumpingYang001:
thanks for reviewing! for omiserver.conf file, I will new another PR, thanks.

jeffaco:
This is probably more involved. I'd suggest leaving that for someone here, unless you're prepared to do all the research without requiring a bunch of support from us.

PR #467, opened 2017-12-05
Add script to dump memory allocation of all OMI processes

palladia:
@Microsoft/omi-devs

### DIFF
--- a/Unix/tools/allocdump
+++ b/Unix/tools/allocdump
@@ -1,0 +1,119 @@
+#!/bin/bash
+
+BINDIR=/opt/omi/bin
+SCRIPTDIR=${BINDIR}/support
+OMISERVER=${BINDIR}/omiserver
+GDB=gdb
+GDB_SCRIPT="$SCRIPTDIR/gdb_allocdump"
+MINIMUM_VERSION="100401000"   # version 1.4.1-0
+CONFIGEDITOR=${BINDIR}/omiconfigeditor
+CONF_FILE=/etc/opt/omi/conf/omiserver.conf
+
+# Verify that the GDB script file exists
+if [ ! -f ${GDB_SCRIPT} ]; then
+    echo "Gdb script file ${GDB_SCRIPT} does not exist"
+    exit 1
+fi
+
+# Verify that gdb exists
+if [ ! -x "$(command -v ${GDB})" ]; then 
+    echo "${GDB} not found in your PATH"
+    exit 1
+fi
+
+# Verify sudo
+EFFECTIVE_ID="$(id -u)"
+SUDO=
+if [ ${EFFECTIVE_ID} -ne 0 ]; then
+    SUDO=sudo
+    if [ ! -x "$(command -v ${SUDO})" ]; then 
+        echo "${SUDO} not found in your PATH"
+        exit 1
+    fi
+    ${SUDO} -v >&1 >/dev/null 
+    if [ $? -ne 0 ]; then
+        echo "You do not have access to run ${SUDO}"
+        exit 1
+    fi
+fi
+
+# Verify that omiserver process is running
+ps -ef | fgrep ${OMISERVER} | fgrep -v fgrep 2>&1 >/dev/null
+if [ $? -ne 0 ]; then
+    echo "Omiserver is not running"
+    exit 1
+fi
+
+# Verify that omiserver.conf is set up properly
+if [ ! -f ${CONF_FILE} ]; then 
+    echo "${CONF_FILE} does not exist"
+    exit 1
+fi
+if [ ! -x ${CONFIGEDITOR} ]; then 
+    echo "${CONFIGEDITOR} does not exist"
+    exit 1
+fi
+for i in INFO DEBUG VERBOSE; do 
+    ${CONFIGEDITOR} loglevel -q $i < ${CONF_FILE} 2>&1 >/dev/null; 
+    [ $? -eq 0 ] && loglevel_okay=1 && break; 
+done
+if [ "$loglevel_okay" != "1" ]; then
+    echo "Please set loglevel to INFO or higher in ${CONF_FILE}"
+    exit 1
+fi
+
+# Verify that omiserver meets minimum version requirement
+VERSION_LINE="$(${OMISERVER} -v)"
+PRODUCT="$(echo $VERSION_LINE | awk -F" |-" '{print $2}')"
+if [ ${PRODUCT} != "OMI" ]; then
+    echo "Incorrect version information in omiserver -v"
+    exit 1
+fi
+VERSION_TEMP="$(echo $VERSION_LINE | awk -F" |-" '{print $3}')"
+VERSION_BUILD="$(echo $VERSION_LINE | awk -F" |-" '{print $4}')"
+VERSION_MAJOR="$(echo $VERSION_TEMP | awk -F. '{print $1}')"
+VERSION_MINOR="$(echo $VERSION_TEMP | awk -F. '{print $2}')"
+VERSION_PATCH="$(echo $VERSION_TEMP | awk -F. '{print $3}')"
+if [ ${#VERSION_MAJOR} -gt 3 \
+     -o ${#VERSION_MINOR} -gt 3 \
+     -o ${#VERSION_PATCH} -gt 2 \
+     -o ${#VERSION_BUILD} -gt 3 ]; then
+    echo "Version number exceeds maximum allowed: ${VERSION_MAJOR}:${VERSION_MINOR}:${VERSION_PATCH}:${VERSION_BUILD}"
+    exit 1
+fi
+VERSION=$((${VERSION_MAJOR} * 100000000 + ${VERSION_MINOR} * 100000 + ${VERSION_PATCH} * 1000 + ${VERSION_BUILD}))
+if [ ${VERSION} -lt ${MINIMUM_VERSION} ]; then
+    echo "Minimum OMI version requirement not met:  ${VERSION_MAJOR}:${VERSION_MINOR}:${VERSION_PATCH}:${VERSION_BUILD}"
+    exit 1
+fi
+
+# Verify that omiserver was built with DEBUG
+echo $VERSION_LINE | fgrep DEBUG 2>&1 >/dev/null
+if [ $? -ne 0 ]; then
+    echo "Not running a DEBUG version of OMI"
+    exit 1
+fi
+
+# Verify that omiserver has symbols
+file ${OMISERVER} | fgrep "not stripped" 2>&1 >/dev/null
+if [ $? -ne 0 ]; then
+    echo "${OMISERVER} does not have symbols"
+    exit 1
+fi
+
+# Verify that we're running on Linux
+if [ "$(uname -s)" != "Linux" ]; then
+    echo "Only Linux platforms records memory allocations"
+    exit 1
+fi
+
+# Invoke gdb script for omiserver and omiengine
+for i in omiserver omiengine
+    do PROCESSID="$(ps -ef | fgrep ${BINDIR}/$i | fgrep -v fgrep | awk '{print $2}')";
+    if [ ! -z "${PROCESSID}" ]; then
+       echo Process id of $i is: ${PROCESSID}
+       $SUDO $GDB -pid ${PROCESSID} -x ${GDB_SCRIPT} >/dev/null 2>&1
+    fi;
+done
+
+exit 0
--- a/Unix/configeditor/configeditor.cpp
+++ b/Unix/configeditor/configeditor.cpp
@@ -451,13 +451,16 @@ int MI_MAIN_CALL main(int argc, const char** argv)
         {
         	std::list<std::string> valueList;
 
-        	_ParseValueList(valueList, value);
-        	if (_IsValueInList(valueList, opts.query) != valueList.end())
-        	{
-        	    cout << prop << " have the value " << opts.query << std::endl;
-        	    return 0;
-        	}
-               cout << prop << " does not have the value " << opts.query << std::endl;
+                if (!fCommentedOut)
+                {
+                    _ParseValueList(valueList, value);
+                    if (_IsValueInList(valueList, opts.query) != valueList.end())
+                    {
+                        cout << prop << " has the value " << opts.query << std::endl;
+                        return 0;
+                    }
+                }
+                cout << prop << " does not have the value " << opts.query << std::endl;
         	return 1;
         }
         else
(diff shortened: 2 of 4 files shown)


Unix/tools/allocdump
```diff
@@ -0,0 +1,32 @@
...
+#   3.  Current user needs to be enabled for sudo
+
+SCRIPTDIR=$(dirname "$0")
+BINDIR=/opt/omi/bin
+GDB=gdb
+SUDO=sudo
+GDB_SCRIPT="$SCRIPTDIR/gdb_allocdump"
+
+if [ ! -f ${GDB_SCRIPT} ]; then
+    echo "Gdb script file ${GDB_SCRIPT} does not exist"
```
jeffaco: Funny abbreviation for `gdb`, suggest lower-casing first letter even though it starts the line.

Unix/tools/allocdump
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
+#
+# Script to run to collect memory allocation information for all running OMI processes
+# Note that in order for this to work:
+#   1.  Omi processes must have been built with debug turned on
```
jeffaco: Hmm. We should create an issue so that, when doing a debug build, a marker file is left around. Or perhaps set up an OMI version file (installinfo or something) that gives the OMI build version and if it's debug or not. Then we can verify that this is the case.

I'd like this script to be as robust as possible so customers know what to do if something goes wrong.

Unix/tools/allocdump
```diff
@@ -0,0 +1,32 @@
...
+#   2.  Omi loglevel must be INFO or higher
```
jeffaco: You can verify this. Can you do so in this script please? If it's not INFO or higher, then print an error and abort rather than just "doing the wrong thing".

Unix/tools/allocdump
```diff
@@ -0,0 +1,32 @@
...
+#   3.  Current user needs to be enabled for sudo
```
jeffaco: You can test this too with something like `sudo /bin/true` and be sure the exit code is zero.

But if the user is currently root, then you don't need this, and SUDO variable should be set to empty (i.e. $SUDO would be ignored). And above message should read `Current user needs to be enabled for sudo, or current user must be running as root user`

Unix/tools/allocdump
```diff
@@ -0,0 +1,32 @@
...
+
+SCRIPTDIR=$(dirname "$0")
+BINDIR=/opt/omi/bin
+GDB=gdb
+SUDO=sudo
```
jeffaco: Note above, this might be conditional based on current effective ID.

Unix/tools/allocdump
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
+#
+# Script to run to collect memory allocation information for all running OMI processes
+# Note that in order for this to work:
+#   1.  Omi processes must have been built with debug turned on
+#   2.  Omi loglevel must be INFO or higher
+#   3.  Current user needs to be enabled for sudo
+
```
jeffaco: Also add:

```
#  4. This script only supports Linux platforms
```

Then verify that you're on a Linux platform and, if you're not, abort with an error.

Unix/tools/allocdump
```diff
@@ -0,0 +1,32 @@
...
+SCRIPTDIR=$(dirname "$0")
+BINDIR=/opt/omi/bin
+GDB=gdb
```
jeffaco: `gdb` may not be installed, test for it and if not found, print an error.

Unix/tools/allocdump
jeffaco: Also note: User may have an alternate `sudo`. I know that `scxcore` has support for this, although I don't think we should check for that. You might want to just print a message up front that `gdb` would be run via `sudo`, if you'll really be doing that (i.e. not running as root user).

Unix/tools/allocdump
```diff
@@ -0,0 +1,32 @@
...
+if [ ! -f ${GDB_SCRIPT} ]; then
+    echo "Gdb script file ${GDB_SCRIPT} does not exist"
+    exit 1
+fi
+
+for i in omiserver omiengine
+    do PROCESSID="$(ps -ef | fgrep ${BINDIR}/$i | fgrep -v fgrep | awk '{print $2}')";
+    if [ ! -z "${PROCESSID}" ]; then
+       echo Process id of $i is: ${PROCESSID}
+       $SUDO $GDB -pid ${PROCESSID} -x ${GDB_SCRIPT} >/dev/null 2>&1
```
jeffaco: Does this work on all platforms, even way back on SuSE-10? Please verify that it does. Otherwise, we may need more smarts here.

Unix/tools/allocdump
```diff
@@ -0,0 +1,32 @@
...
+    fi;
+done
+
+# for omiagent(s)
+ps -ef | fgrep ${BINDIR}/omiagent | fgrep -v fgrep | awk '{print $2}' | while read -r PROCESSID; do
+    echo Process id of omiagent is: ${PROCESSID}
+    $SUDO $GDB -pid ${PROCESSID} -x ${GDB_SCRIPT} >/dev/null 2>&1
```
jeffaco: This is going to be sucky in providers without the issue implemented to put providers into debug mode. Sure, you know a provider is leaking. But determining WHICH provider will be very tricky indeed.

Unix/tools/allocdump
```diff
@@ -0,0 +1,123 @@
+#!/bin/bash
+
+SCRIPTDIR=$(dirname "$0")
```
jeffaco: The `dirname` command isn't necessarily reliable to use for other paths (like on line 7, for `gdb_allocdump`. Although, in this case, it's probably okay.

Why not just hard-code where `installbuilder` will put this, thus playing it "safe"?

Unix/tools/allocdump
```diff
@@ -0,0 +1,123 @@
...
+EFFECTIVE_ID="$(id -u)"
+SUDO=
+if [ ${EFFECTIVE_ID} -ne 0 ]; then
+    SUDO=sudo
+    if [ ! -x "$(command -v ${SUDO})" ]; then 
+        echo "${SUDO} not found in your PATH"
+        exit 1
+    fi
+    ${SUDO} -v >&1 >/dev/null 
+    if [ $? -eq 1 ]; then
```
jeffaco: Hmm. This is fine for a "blanket" `/etc/sudoers` file (i.e. you have perms to do anything). But what about restricted commands? If `/etc/sudoers` is restricted to certain commands, you can update your credentials, but that doesn't mean you can actually run this script.

Not sure how important this is. I guess, if you catch the error and print something meaningful, this is fine.

Unix/tools/allocdump
```diff
@@ -0,0 +1,123 @@
...
+fi
+if [ ! -x ${CONFIGEDITOR} ]; then 
+    echo "${CONFIGEDITOR} does not exist"
+    exit 1
+fi
+for i in INFO DEBUG VERBOSE; do 
+    ${CONFIGEDITOR} loglevel -q $i < ${CONF_FILE} 2>&1 >/dev/null; 
+    [ $? -eq 0 ] && found=1 && break; 
+done
+if [ "$found" != "1" ]; then
```
jeffaco: `found` isn't initialized. That's fine, but perhaps this should be a little more unique, like `loglevel_okay`?

Unix/tools/allocdump
```diff
@@ -0,0 +1,123 @@
...
+VERSION_TEMP="$(echo $VERSION_LINE | awk -F" |-" '{print $3}')"
+VERSION_BUILD="$(echo $VERSION_LINE | awk -F" |-" '{print $4}')"
+VERSION_MAJOR="$(echo $VERSION_TEMP | awk -F. '{print $1}')"
+VERSION_MINOR="$(echo $VERSION_TEMP | awk -F. '{print $2}')"
+VERSION_PATCH="$(echo $VERSION_TEMP | awk -F. '{print $3}')"
+if [ ${#VERSION_MAJOR} -gt 3 \
+     -o ${#VERSION_MINOR} -gt 3 \
+     -o ${#VERSION_PATCH} -gt 2 \
+     -o ${#VERSION_BUILD} -gt 3 ]; then
+    echo "Invalid version number: ${VERSION_MAJOR}:${VERSION_MINOR}:${VERSION_PATCH}:${VERSION_BUILD}"
```
jeffaco: That's not exactly an "invalid version number", it's a version number where a field is too long. Make error message more meaningful please.

Unix/tools/allocdump
```diff
@@ -0,0 +1,123 @@
...
+
+# Verify that omiserver has symbols
+file ${OMISERVER} | fgrep "not stripped" 2>&1 >/dev/null
+if [ $? -ne 0 ]; then
+    echo "${OMISERVER} does not have symbols"
+    exit 1
+fi
+
+# Verify that we're running on Linux
+if [ "$(expr substr $(uname -s) 1 5)" != "Linux" ]; then
```
jeffaco: When is `uname -s` != "Linux" on Linux systems? I checked RedHat, CentOS, and Ubuntu, and `uname -s` == "Linux" on all of that. In other cases, we do a straight string compare, could you do that here?

Unix/tools/allocdump
```diff
@@ -0,0 +1,123 @@
...
+for i in omiserver omiengine
+    do PROCESSID="$(ps -ef | fgrep ${BINDIR}/$i | fgrep -v fgrep | awk '{print $2}')";
+    if [ ! -z "${PROCESSID}" ]; then
+       echo Process id of $i is: ${PROCESSID}
+       $SUDO $GDB -pid ${PROCESSID} -x ${GDB_SCRIPT} >/dev/null 2>&1
+    fi;
+done
+
+# for omiagent(s)
+ps -ef | fgrep ${BINDIR}/omiagent | fgrep -v fgrep | awk '{print $2}' | while read -r PROCESSID; do
```
jeffaco: Is it worthwhile to look at OMI-specific leaks in omiagent at all? If there was a leak there, would we see that with existing debugging aids?

Unix/tools/allocdump
palladia: Switching to hard-coded.

Unix/tools/allocdump
palladia: No change.

Unix/tools/allocdump
palladia: Changed name to loglevel_okay.

Unix/tools/allocdump
palladia: Changed error message to "Version number exceeds maximum allowed".

Unix/tools/allocdump
palladia: Saw this in an online link.  Switch to just straight comparison.

Unix/tools/allocdump
palladia: Disabled omiagent check.

Unix/tools/allocdump
jeffaco: How about changing line 34 to read:

```
if [ $? -ne 0 ]; then
```

Seems safer ...

Unix/tools/allocdump
```diff
@@ -0,0 +1,123 @@
...
+    ${SUDO} -v >&1 >/dev/null 
+    if [ $? -eq 1 ]; then
+        echo "You do not have access to run ${SUDO}"
+        exit 1
+    fi
+fi
+
+# Verify that omiserver process is running
+ps -ef | fgrep ${OMISERVER} | fgrep -v fgrep 2>&1 >/dev/null
+if [ $? -eq 1 ]; then
```
jeffaco: You did this here too. In general, I like to check for success as `$? -eq 0` and error as `$? -ne 0`.

The issue is that error status can be anything that's non-zero (except in very specific cases with sub-processes and and `wait` operations). Most scripts and programs, for examples, will return 0 for success and non-zero for failure (where the specific error code reflects the type of error that occurred).

Please fix this and any other occurences.

Unix/tools/allocdump
```diff
@@ -0,0 +1,123 @@
...
+fi
+VERSION=$((${VERSION_MAJOR} * 100000000 + ${VERSION_MINOR} * 100000 + ${VERSION_PATCH} * 1000 + ${VERSION_BUILD}))
+if [ ${VERSION} -lt ${MINIMUM_VERSION} ]; then
+    echo "Minimum OMI version requirement not met:  ${VERSION_MAJOR}:${VERSION_MINOR}:${VERSION_PATCH}:${VERSION_BUILD}"
+    exit 1
+fi
+
+# Verify that omiserver was built with DEBUG
+echo $VERSION_LINE | fgrep DEBUG 2>&1 >/dev/null
+if [ $? -ne 0 ]; then
```
jeffaco: Yeah, like that! 👍

Unix/tools/allocdump
jeffaco: Did you verify that this works on SuSE 10? I didn't see a comment on this ... for this, I'll want another iteration, sorry. You're getting very close, though!

Unix/tools/allocdump
```diff
@@ -0,0 +1,123 @@
...
+# Invoke gdb script for omiserver and omiengine
+for i in omiserver omiengine
+    do PROCESSID="$(ps -ef | fgrep ${BINDIR}/$i | fgrep -v fgrep | awk '{print $2}')";
+    if [ ! -z "${PROCESSID}" ]; then
+       echo Process id of $i is: ${PROCESSID}
+       $SUDO $GDB -pid ${PROCESSID} -x ${GDB_SCRIPT} >/dev/null 2>&1
+    fi;
+done
+
+# for omiagent(s) - disabled for now
```
jeffaco: Do we want to keep disabled code? It's close enough to the above: Why not just nuke it?

Unix/tools/allocdump
```diff
@@ -0,0 +1,123 @@
...
+#ps -ef | fgrep ${BINDIR}/omiagent | fgrep -v fgrep | awk '{print $2}' | while read -r PROCESSID; do
+#    echo Process id of omiagent is: ${PROCESSID}
+#    $SUDO $GDB -pid ${PROCESSID} -x ${GDB_SCRIPT} >/dev/null 2>&1
+#done
```
jeffaco: As is, exit status will be whatever line 115 is, depending on default of error controls here.

How about an explicit success exit (i.e. `exit 0`)?

Unix/tools/allocdump
palladia: Verified on SuSE 10.

PR #468, opened 2017-12-06
Check that system keytab exists in crontab

palladia:


palladia:
microsoft/omi-devs

JumpingYang001:
If systab exist, but ktutil doesn't install it will also get error in log.

JumpingYang001:
in normal case, if systab exist ktutil should exist and work?

yakman2020:
ktutil is a standard part of the system. Its like asking what happens in "ls" isn't installed.

PR #471, opened 2017-12-19
fix ssl link issue in installssllinks

### LINKED ISSUE #470: installssllinks script should be more robust
If `/usr/bin/openssl` is linked statically to SSL, we'll get output such as the following when doing `ldd` against it:

```
[armotz@wemgaaccpub001 ~]$ ldd /bin/openssl
        linux-vdso.so.1 =>  (0x00007fff84bc9000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007efd4bf85000)
        libc.so.6 => /lib64/libc.so.6 (0x00007efd4bbc2000)
        /lib64/ld-linux-x86-64.so.2 (0x00007efd4c193000)
```

This causes this line in `installssllinks`:

```
LIBSSL_PATH=`ldd ${OPENSSL_PATH} | grep libssl.so | awk '{print $3}'`
```

to return an empty string (with no error so the next line doesn't catch the problem).

Error handling should be enhanced to catch this and abort with a meaningful error.

JumpingYang001:
fix #470 , @Microsoft/omi-devs , could you help to review this PR? thanks very much!

### DIFF
--- a/Unix/scripts/installssllinks
+++ b/Unix/scripts/installssllinks
@@ -49,12 +49,12 @@ create_ssl_links() {
 
     OPENSSL_PATH=`which openssl`
     LIBSSL_PATH=`ldd ${OPENSSL_PATH} | grep libssl.so | awk '{print $3}'`
-    if [ $? -ne 0 ]; then
+    if [ $? -ne 0 ] || [ ! -e "${LIBSSL_PATH}" ]; then
         echo "Error: Unable to determine libssl.so path" >&2
         exit 2
     fi
     LIBCRYPTO_PATH=`ldd ${OPENSSL_PATH} | grep libcrypto.so | awk '{print $3}'`
-    if [ $? -ne 0 ]; then
+    if [ $? -ne 0 ] || [ ! -e "${LIBCRYPTO_PATH}" ]; then
         echo "Error: Unable to determine libcrypto.so path" >&2
         exit 2
     fi


JumpingYang001:
just new pr first, I will verify the fix on box today, thanks.

JumpingYang001:
@jeffaco  , I have fixed them.
Here is the result before fix:
```
root@omi-ub16x64-t1:~/openssl-1.0.2n# ./config --prefix=/usr --openssldir=/etc/ssl --libdir=lib no-shared
root@omi-ub16x64-t1:~/openssl-1.0.2n# make INSTALL_PREFIX=/ssltest install
root@omi-ub16x64-t1:~/openssl-1.0.2n# ldd /ssltest/usr/bin/openssl                                                                                                                                                      
        linux-vdso.so.1 =>  (0x00007ffff77f0000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f79a4491000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f79a40c7000)
        /lib64/ld-linux-x86-64.so.2 (0x000055b50afc8000)
root@omi-ub16x64-t1:~/openssl-1.0.2n# mv /usr/bin/openssl /usr/bin/openssl2
root@omi-ub16x64-t1:~/openssl-1.0.2n# ln -s /ssltest/usr/bin/openssl /usr/bin/openssl                                                                                                                                   
root@omi-ub16x64-t1:~/openssl-1.0.2n# /opt/omi/bin/support/installssllinks                                                                                                                              
dirname: missing operand
Try 'dirname --help' for more information.
basename: missing operand
Try 'basename --help' for more information.
dirname: missing operand
Try 'dirname --help' for more information.
basename: missing operand
Try 'basename --help' for more information.
```
Here is result after fixing:
```
root@omi-ub16x64-t1:~/openssl-1.0.2n# /opt/omi/bin/support/installssllinks                                                                                                                                                   
Error: Unable to determine libssl.so path
```

JumpingYang001:
@jeffaco , do you mean like following multi-lines?
```
     if [ $? -ne 0 ] ; then
          echo "Error: Unable to determine libssl.so path" >&2
          exit 2
      fi
      if [ !-e ${LIBSSL_PATH} ]; then
          echo "Error: Could not find libssl.so in openssl" >&2
          exit 2
      fi
```

jeffaco:
No. I want a multi-line commit message. Like with the line that you already have, and then the rest of the lines that I gave you. [Here](https://github.com/Microsoft/omi/commit/2fa4efe37e1146b7c67ea8afc25e4d23c8a0bce3) is an example of a multi-line commit message.

JumpingYang001:
ho, I see, thank you!
I will update the commit message. :)

JumpingYang001:
@jeffaco , how about current commit message?

jeffaco:
Nope. I gave you the exact commit message I wanted, please use that.  Here's the complete commit message I'm looking for (adjusted for line length):

```
Be sure that libssl.so/libcrypto.so are resolved.                               

It's possible, if SSL is linked statically to /bin/openssl, that
libssl.so and/or libcrypto.so won't be found, which will cause
the script to fail in various ways.

Verify that the libraries are actually found or give a meaningful
error if not.
```

JumpingYang001:
@jeffaco , the commit message is updated, thanks.

JumpingYang001:
I need to update the bash shell script.

JumpingYang001:
@jeffaco , I have updated the commit message and correct the shell script to work on both with/without libssl.so loaded.

JumpingYang001:
thanks for reviewing!

PR #472, opened 2017-12-22
George/debugassert

palladia:
1.  Removed unnecessary DEBUG_ASSERT in protocol.c.
2.  Modified installbuilder/GNUMakefile so that enable-debug option of configure will result with just one kit (with debug in its name).

### DIFF
--- a/Unix/installbuilder/GNUmakefile
+++ b/Unix/installbuilder/GNUmakefile
@@ -59,19 +59,25 @@ DATAFILES += SunOS_5.11.data
  endif
 endif
 
+ifeq ($(ENABLE_DEBUG),1)
+  WITH_SYMBOLS=debug
+else
+  WITH_SYMBOLS=withsymbols
+endif
+
 # New output format for OMI on Linux (avoids ugly renaming since we always build universal)
 ifeq ($(PF),Linux)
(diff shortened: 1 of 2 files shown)


Unix/installbuilder/GNUmakefile
```diff
@@ -200,16 +214,19 @@ ifneq ($(PF),Darwin)
 		--RELEASE=$(CONFIG_PATCH_LEVEL) \
 		$(OUTPUTFILE_LINE_WITHSYMBOLS) \
 		$(SSL_BUILD_LINE) \
-		--DPKG_LOCATION=$(TOP)/../../pal/installer/InstallBuilder/tools/bin/dpkg-deb-$(PF_ARCH) \
+		$(DPKG_LOCATION) \
 		--DATAFILE_PATH=$(TOP)/installbuilder/datafiles \
 		$(DATAFILES) Linux_DPKG.data
     endif # ifeq ($(BUILD_DPKG),1)
 
+    ifneq ($(ENABLE_DEBUG),1)
```
jeffaco: This one conditional can be combined with conditionals below. Or, said another way, if we're not DEBUG, then do lines 228-275. Or something like that.

palladia: Done

PR #473, opened 2017-12-28
Minor fixes (proposal)

kant:


jeffaco:
Merged with a better description in commit message.

PR #474, opened 2018-01-04
Improvements from debugging oms system:

palladia:
1. Fixed omiserver memory leak in handling CreateAgentReq
2. Improvements to logfile to distinguish server from engine output
3. Changed missing NTLM message from FATAL to WARNING
4. Modified DumpAllocList() to print address of interest instead of header
5. Removed memory leak of krb5KeytabPath in server
6. Removed trace statements in signal handlers
7. Fixed warning-level log message in Engine's handling of BinNotification msg

### DIFF
--- a/Unix/pal/alloc.c
+++ b/Unix/pal/alloc.c
@@ -124,7 +124,9 @@ void PAL_DumpAllocList()
 
     for (p = _list; p; p = p->next)
     {
-        trace_DumpAllocList_Block(p->file, (int)p->line, p, p->magic, p->id, p->size);
+        // The address of interest is not the header, hence we print p+1 to 
+        // skip the header
+        trace_DumpAllocList_Block(p->file, (int)p->line, p+1, p->magic, p->id, p->size);
     }
 }
 
--- a/Unix/protocol/protocol.c
+++ b/Unix/protocol/protocol.c
@@ -65,6 +65,7 @@ static char s_secretString[S_SECRET_STRING_LENGTH];
 static HashMap s_protocolSocketTracker;
 static Lock s_trackerLock;
 static MI_Result (*authenticateCallback)(PamCheckUserResp*);
+static char s_type = 'U';  // 'U': unknown, 'E': engine, 'S': server
 
 /*
 **==============================================================================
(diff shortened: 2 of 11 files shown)


palladia:
@Microsoft/omi-devs

Unix/pal/alloc.c
```diff
@@ -124,7 +124,7 @@ void PAL_DumpAllocList()
 
     for (p = _list; p; p = p->next)
     {
-        trace_DumpAllocList_Block(p->file, (int)p->line, p, p->magic, p->id, p->size);
+        trace_DumpAllocList_Block(p->file, (int)p->line, p+1, p->magic, p->id, p->size);
```
jeffaco: Um, why `p+1`? Minimally, this needs a comment. But it looks like an error, explain please.

Unix/protocol/protocol.c
```diff
@@ -2067,30 +2072,27 @@ static Protocol_CallbackResult _ProcessReceivedMessage(
...
                         handler = newHandler;
 
-                        if(_SendAuthResponse(handler, binMsg->result, binMsg->authFile, forwardSock, 
-                                             binMsg->uid, binMsg->gid))
-                        {
-                            ret = PRT_CONTINUE;
-                        }
+                        if(!_SendAuthResponse(handler, binMsg->result, binMsg->authFile, forwardSock, 
+                                              binMsg->uid, binMsg->gid))
+                            trace_ClientAuthResponseFailed();
```
jeffaco: Does OMI have a coding standard where all `if` statements should have surrounding braces?

Unix/pal/alloc.c
palladia: Added comment to explain +1.

Unix/protocol/protocol.c
palladia: Added braces.

Unix/pal/alloc.c
```diff
@@ -124,7 +124,9 @@ void PAL_DumpAllocList()
 
     for (p = _list; p; p = p->next)
     {
-        trace_DumpAllocList_Block(p->file, (int)p->line, p, p->magic, p->id, p->size);
+        // The address of interest is not the header, hence we print p+1 to
```
jeffaco: s/print/use/:

```
// The address of interest is not the header, hence we use p+1 to
```

palladia:
@paulcallen - please review.

PR #475, opened 2018-01-08
Remove ENCRYPT_DECRYPYT flag from build tool.

jeffaco:
Misspelled, and not needed anyway since it's in config.h.


jeffaco:

PR #480, opened 2018-01-17
Remove WINDOWS and MSC code

palladia:


Unix/base/user.c
```diff
@@ -41,84 +40,6 @@
...
-{
-    MI_UNUSED(uid);
-    MI_UNUSED(size);
-    MI_UNUSED(path);
-    MI_UNUSED(content);
-
-    return -1;
-}
-
-#elif defined(CONFIG_POSIX)
```
paulcallen: based on our comment should this #elif not be a $if now?

paulcallen: and where is the #endif?

Unix/codec/mof/parser/utility.c
```diff
@@ -1092,13 +1023,8 @@ void ReportFinalError(
     switch (id)
     {
     case ID_OUT_OF_MEMORY:
-#if defined(_MSC_VER)
-        errorCode = (MI_Uint32)E_OUTOFMEMORY;
-        errorType = MI_RESULT_TYPE_HRESULT;
-#else
         errorCode = ENOMEM;
         errorType = MI_T("STDC"); /* todo: Fixup error type for non-win */
```
paulcallen: um, I believe this should be MI_RESULT_TYPE_ERRNO

Unix/base/user.c
palladia: I believe this is ok.  No change.

palladia: Also ok.

Unix/codec/mof/parser/utility.c
palladia: Changed.

PR #484, opened 2018-01-22
Fixes failed restart on Redhat6 and Cent6

palladia:


Unix/installbuilder/service_scripts/omid.upstart
```diff
@@ -18,11 +18,9 @@
 
 description	"Microsoft OMI Server"
 
-start on (starting network-interface
-      or starting networking)
+start on runlevel [2345]
```
jeffaco: So the problem here is that this will only be triggered on runlevel changes (or booting). Based on timing, we could come up before the network comes up, and that would cause us to fail startup.

Is the problem that the network is called something else (service name wise) on RHEL? Is there a related name on both systems that we may be able to use?

If I stopped the network, I'd expect OMI to run into trouble. And then it would try to restart, most likely, causing problems.

PR #486, opened 2018-02-01
Fix reload_dispatcher file as an alternate mechanism to reload

palladia:


JumpingYang001:
@palladia , how to test this change? I don't understand the change, can you help to provide some scenarios for this? thank you!

PR #487, opened 2018-02-06
Add additional logging at OMI startup

palladia:
At OMI startup, error conditions are logged to stderr, but not logged to file.  This PR adds additional file logging if the log file is ready.

### DIFF
--- a/Unix/base/oiomi.h
+++ b/Unix/base/oiomi.h
@@ -53,6 +53,156 @@ FILE_EVENT0(10005, trace_EngineProcessTerminated_Impl, LOG_CRIT, PAL_T("abnormal
 #endif
 FILE_EVENT1(10006, trace_CriticalError_Impl, LOG_CRIT, PAL_T("%T"), const TChar*)
 #if defined(CONFIG_ENABLE_DEBUG)
+#define trace_OMI_Not_Running() trace_OMI_Not_Running_Impl(__FILE__, __LINE__)
+#else
+#define trace_OMI_Not_Running() trace_OMI_Not_Running_Impl(0, 0)
+#endif
+FILE_EVENT0(10007, trace_OMI_Not_Running_Impl, LOG_CRIT, PAL_T("OMI is not running"))
+#if defined(CONFIG_ENABLE_DEBUG)
+#define trace_OMI_Already_Running() trace_OMI_Already_Running_Impl(__FILE__, __LINE__)
+#else
+#define trace_OMI_Already_Running() trace_OMI_Already_Running_Impl(0, 0)
+#endif
+FILE_EVENT0(10008, trace_OMI_Already_Running_Impl, LOG_CRIT, PAL_T("OMI is already running"))
+#if defined(CONFIG_ENABLE_DEBUG)
+#define trace_Stop_OMI_Failed() trace_Stop_OMI_Failed_Impl(__FILE__, __LINE__)
+#else
+#define trace_Stop_OMI_Failed() trace_Stop_OMI_Failed_Impl(0, 0)
+#endif
+FILE_EVENT0(10009, trace_Stop_OMI_Failed_Impl, LOG_CRIT, PAL_T("Failed to stop"))
+#if defined(CONFIG_ENABLE_DEBUG)
+#define trace_ReloadConfig_OMI_Failed() trace_ReloadConfig_OMI_Failed_Impl(__FILE__, __LINE__)
+#else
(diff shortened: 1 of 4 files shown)


Unix/base/oiomi.h
```diff
@@ -2237,6 +2387,42 @@ FILE_EVENT3(30213, trace_Selector_AddHandler_AlreadyThere_Impl, LOG_WARNING, PAL
 #endif
 FILE_EVENT3(30214, trace_Selector_RemoveHandler_NotThere_Impl, LOG_WARNING, PAL_T("Selector_RemoveHandler: selector=%p, handler=%p, name=%T NOT REGISTERED"), Selector *, Handler *, const TChar *)
 #if defined(CONFIG_ENABLE_DEBUG)
+#define trace_Stop_OMI() trace_Stop_OMI_Impl(__FILE__, __LINE__)
```
JumpingYang001: I have a question about WARNING logs here, if someone stop/restart omi, it will log a WARNING, is it? will this be confused by user? just pick it up, thanks.

palladia: Since starting and stopping is not a common event, wouldn't it make sense to log as WARNING?

Currently, we have no record of someone starting or stopping OMI, since default log level is WARNING.  

We have a new issue where customers are complaining that OMI didn't restart.  We don't even have any record in our logs that OMI restart was attempted.

JumpingYang001: ok, hi @jeffaco , does WARNING make sense here?

jeffaco: George and I discussed it. Note that we have precedence here, as [SCXCore](https://github.com/Microsoft/SCXcore) logs agent startup and shutdown in it's log files (it also logs version information). In this way, you have exact startup/shutdown times for the server.

Note that, while we see a lot of startup/shutdown events in our testing, in a normal customer case OMI would start at system boot and would run indefinitely until machine shutdown or OMI upgrade. Thus, regular customers should not see very many of these messages logged.

Ultimately, for diagnostic purposes an an unrelated issue, we wanted our logfile to show startup/shutdown events. Since the default log level is WARNING, these messages had to be WARNING as well to make it though the minimum log level filter.

Let me know if you have further questions.

Unix/base/oiomi.h
```diff
@@ -53,6 +53,156 @@ FILE_EVENT0(10005, trace_EngineProcessTerminated_Impl, LOG_CRIT, PAL_T("abnormal
...
+#endif
+FILE_EVENT0(10030, trace_Failed_Set_User_Engine_Impl, LOG_CRIT, PAL_T("Failed to set uid/gid of engine"))
+#if defined(CONFIG_ENABLE_DEBUG)
+#define trace_Failed_Launch_Engine() trace_Failed_Launch_Engine_Impl(__FILE__, __LINE__)
+#else
+#define trace_Failed_Launch_Engine() trace_Failed_Launch_Engine_Impl(0, 0)
+#endif
+FILE_EVENT0(10031, trace_Failed_Launch_Engine_Impl, LOG_CRIT, PAL_T("Failed to launch engine"))
+#if defined(CONFIG_ENABLE_DEBUG)
 #define trace_AgentMgrHandleRequestForNS(a0) trace_AgentMgrHandleRequestForNS_Impl(__FILE__, __LINE__, tcs(a0))
```
jeffaco: It's odd that the next message isn't an error, which is what's in the source file. What will this do for our ability to upgrade while maintaining message IDs?

jeffaco:

Unix/base/oiomi.h
JumpingYang001: ok, got it, thanks.

Unix/base/oiomi.h
JumpingYang001: I saw LOG_CRIT is 10000-10006, LOG_ERR is 20000-20148, LOG_WARNING is 30000-30213, LOG_INFO is 40000-40043, LOG_DEBUG is 45000-45380, LOG_VERBOSE is 55000-55056 and 60000-60024,
then a 60000-60017 for LOG_DEBUG at the end.
Duplicated ids are from 60000-60017 now, they used different FILE_EVENT_number function.

jeffaco: This was an error on my part; at quick glance, I hadn't realized that the GitHub UI was not showing me contiguous lines.

palladia: I am moving Bruce's auth-related trace messages from 60000-block to 50000-block.  This avoids conflict, but breaks backward compatibility a bit.  Can't do anything about that.

PR #490, opened 2018-02-20
Add option to enable a provider to run in its own agent

palladia:


Unix/etc/omiserver.conf
```diff
@@ -72,3 +72,6 @@
 ## if USER is specified within <> as the service account, 
 ##      it will be replaced with $USER when copied to output*
 service=<USER>
+
+## Enabling this will cause each provider to run under its own agent
+#individualAgent=false
```
jeffaco: This isn't a good name. This is more a debugging feature to help debug provider crashes in omiagent. How about something like `agentDebugging`?

I'd also like to see a much better description for this. Perhaps something like:

```
# Enabling this will cause each provider to run under it's own omiagent
# process. This will take considerably more memory, but is useful for
# diagnosing omiagent cores due to a provider fault. Setting of `true`
# means that each provider runs under it's own omiagent process.
#agentDebugging=false

File: Unix/installbuilder/conf/omiserver.conf [hunk 2]
```diff
@@ -51,3 +51,6 @@ httpsport=0
 ##
 #NoSSLv2=true
 #NoSSLv3=false
+
+## Enabling this will cause each provider to run under its own agent
+#individualAgent=false
```
jeffaco: See above.

Unix/etc/omiserver.conf
palladia: I wasn't crazy about my choice of name anyway, so agentDebugging it is!

palladia: Adding Paul to reviewer list.

PR #492, opened 2018-03-06
OMI does not install in Docker

palladia:


JumpingYang001:
@palladia Hi Geroge , so our fix is don't use `systemctl` anymore? since `service` will redirect to `systemctl` on centos.

PR #494, opened 2018-03-13
Add access denied error message for reg files

palladia:
Customers were getting corrupted-file messages if omiengine was not able to read the provider registration files.


Unix/base/oi_traces.h
```diff
@@ -426,6 +426,9 @@ void trace_HTTP_ClientAuthFailed(const char * major, const char * minor);
 OI_EVENT("Failed to send auth response to client")
 void trace_ClientAuthResponseFailed();
 
+OI_EVENT("Reg file %s access denied. It will be skipped by the server")
```
jeffaco: Make that `Registration file` ...

PR #500, opened 2018-03-27
fix ptr/char comparaison

aaptel:
dereference pointer before comparing.


palladia:
What is the problem this fix attempts to address?

The fix seems reasonable, but Eric, please review.

aaptel:
The old code was wrong and it was breaking the build with GCC7 and and right warnings/errors flags.

PR #501, opened 2018-03-28
Add SSL cert checking options

JumpingYang001:
@paulcallen , I add 4 new options for omicli "-skipselfsigncheck, -skipcacheck, -skipcncheck or -skiprevocationcheck", could you help to review this PR? thank you very much!

PowerShell only contains 3 options for https remoting: "-SkipCACheck,-SkipCNCheck,-SkipRevocationCheck", and "-skipselfsigncheck" is a new option that Pentest pointed out for notification of Self-Signed Certs.

### DIFF
--- a/Unix/cli/cli_c.c
+++ b/Unix/cli/cli_c.c
@@ -71,6 +71,10 @@ struct Options
     MI_Boolean xml;
     MI_Uint32 maxEnvSize;
     MI_Uint32 maxElements;
+    MI_Boolean skipselfsigncheck; 
+    MI_Boolean skipcacheck;
+    MI_Boolean skipcncheck;
+    MI_Boolean skiprevocationcheck;
 };
 
 static struct Options opts;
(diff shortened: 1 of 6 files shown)


palladia:
New directive from Li is that low/medium PenTest items should only be checked in if its risks/rewards ratio is low.  Paul, this item is considered "medium" by PenTest Team, is this a worthwhile item to check in?

JumpingYang001:
@paulcallen how do you think?

JumpingYang001:
**self-signed cert**
The medium Pentest issue is about self-signed cert , and the self-signed cert should be ok for OMS or scx(they don't expose the cert to public, OMS call local binary protocol, and scx for OM are on-premised), and for PSRP, it might be exposing to public, when we say self-signed cert, we are considering it is a Linux/Unix server(Mac don't support server now), and when user are remoting from Windows to Linux, it really like the doc say it will not be notified about the remoting server is a self-signed cert on Windows side.

For 3 options related PSRP, X509_V_ERR_HOSTNAME_MISMATCH is only supported by openssl 1.1.0 that is related to -skipcncheck, so it will fail for openssl less than 1.1.0, I don't have confident to make this PR go into product, and that doesn't related to the Pentest issue, after a long thinking, I want to close this PR, thanks.

JumpingYang001:
Close the PR, and will create new one if we really need one fix.

Issue #502, opened 2018-03-28
segfault in InteractionProtocolHandler_Session_Connect in debug builds on failed connections.


dantraMSFT:
Connecting through PowerShell/PSRP using basic auth with invalid credentials causes a segfault in debug builds.

The problem is due to an invalid pointer dereference.

At the lowest level, when the error is posted, PostResult has the following lines at the end:

Message_AddRef(&errorMsg->base);
Strand_ScheduleAux(&self->strand, PROTOCOLSOCKET_STRANDAUX_POSTMSG);
PostResultMsg_Release(errorMsg);

After calling Strand_ScheduleAux, the operation that was passed into InteractionProtocolHandler_Session_Connect has been freed (in debug builds, the memory block is filled with 0xdddddddddddddddd). This causes the seg fault for the following reason:

1: MI_RESULT_ACCESS_DEFINED gets returned to WsmanClient_New_Connector

2: WsmanClient_New_Connector jumps to finished2 and posts another error result but then returns MI_RESULT_OK to the caller.

3: InteractionProtocolHandler_Session_Connect receives the MI_RESULT_OK and attempts to continue but it is referencing a stale memory pointer; it checks for a non-null pointer in one of the fields. In release builds, this check succeeds, the memory block is zero filled. In debug builds, the field is non-null (0xdddddddddddddddd) and the pointer is dereferenced causing a segfault.

Returning MI_RESULT_OK appears to be suspect since it requires dereferencing a stale memory pointer.  Alternatively, the operation is getting deleted when it should not be.

JumpingYang001:
I will check it today.

JumpingYang001:
we repro the issue today, will continue to investigate it tomorrow.

JumpingYang001:
I have fixed the previous issue cause this issue, and now it will never hit or go into this issue, so just close it.

JumpingYang001:
https://github.com/PowerShell/psl-omi-provider/issues/117 can repro this issue, will fix it soon.

JumpingYang001:
Found the root cause now:

**debug stack:**
```
InteractionProtocolHandler_Operation_Strand_Finish (self_=0x7f77940d35d8) at InteractionProtocolHandler.c:561
561         PAL_Free(operation);
(gdb) p operation->protocolConnection
$23 = (InteractionProtocolHandler_ProtocolConnection *) 0x7f77940d35c8
(gdb) s
[...]

PR #504, opened 2018-04-04
Disable ENABLECLEARING to avoid crash

JumpingYang001:
Fix issue #502 , @palladia  @paulcallen , could you help to review it? thanks.

### DIFF
--- a/Unix/pal/alloc.c
+++ b/Unix/pal/alloc.c
@@ -18,7 +18,8 @@
 #if defined(USE_ALLOCATOR)
 # define ALLOC_MAGIC 0x89ABCDEF
 # define ENABLE_FILLING
-# define ENABLE_CLEARING
+// disable ENABLE_CLEARING to avoid crash for debug build
+// # define ENABLE_CLEARING
 
 static AllocStats _stats;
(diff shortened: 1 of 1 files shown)


dantraMSFT:
@JumpingYang001 This does not fix the underlying problem of the operation instance being deleted before returning.  As I mentioned in when I reactivated the issue, In release builds, I'm seeing a process failure due to a double free.   

Removing the ENABLE_CLEARING flag doesn't fix the root cause, that the call chain within Strand_ScheduleAux is deleting a pointer that it doesn't own.

In debug builds, this problem is caught in InteractionProtocolHandler_Session_Connect which attempts to use the freed memory.  In release builds, the problem is caught when the owner of the memory attempts to free it resulting in a double free.

This is exactly what was happening the first time around and this PR (https://github.com/Microsoft/omi/commit/ba1958ec946fc7705550d4d78796d5719e54c2cd) fixed that case.
However, I'm hitting another error path, server not reachable, that is not fixed and there may be others.

JumpingYang001:
Just close this PR, will open another PR to fix the release build issue  #502

PR #506, opened 2018-04-10
Fix and improve omid.upstart

JumpingYang001:
PCF customer meet a upstart issue on Ubuntu14.04, and there are many network creating/deleting as following, I have fixed and verified it works on customer's box, also verified on Ubuntu12 and Ubuntu14, we only call initctl start/stop/restart omid on these 2 platforms.

```
network-interface (s-010255079004) start/running

network-interface (s-010255079013) start/running

network-interface (s-010255079031) start/running

network-interface (s-010255079005) start/running

network-interface (s-010255079012) start/running

network-interface (s-010255079020) start/running

...
```

@palladia @paulcallen , could you help to review my fixes? thanks

### DIFF
--- a/Unix/installbuilder/service_scripts/omid.upstart
+++ b/Unix/installbuilder/service_scripts/omid.upstart
@@ -18,10 +18,10 @@
 
 description	"Microsoft OMI Server"
 
-start on (starting network-interface
+start on (starting network-interface INTERFACE=lo
       or starting networking)
 
-stop  on (stopping network-interface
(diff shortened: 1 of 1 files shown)


JumpingYang001:
To repro the issue with exist release build: create 3~5 virtual network on Azure box, and run "initctl stop network-interface INTERFACE=eth1" and "initctl start network-interface INTERFACE=eth1", you will see it trigger omi to start/stop no matter omi is active or not, and do these things several times in same time, you will get omi go into strange/struggle issue.

PR #509, opened 2018-04-13
Avoid deadlock in Engine if Server dies

palladia:
If server dies unexpectedly, the lock in agentmgr could be locked twice.  Since our locks are not recursive, this second locking will cause engine to hang forever.


palladia:
@Microsoft/omi-devs

JumpingYang001:
@palladia , how to repro the deadlock?

palladia:
@JumpingYang001 The only way for me to repro this is to use gdb and stop omiengine  on _CreateAgent(), kill omiserver manually, and continue from debugger.

PR #515, opened 2018-05-02
Check for ktutil

palladia:


JumpingYang001:
In this case, it means that user have /etc/krb5.keytab, but don't have ktutil, that's an abnormal case, the user might touch the file manually that cause this scenario.

PR #516, opened 2018-05-03
Ensure that no email is generated by cron

palladia:


JumpingYang001:
what's difference between `>/dev/null 2>&1` and `2>&1 >/dev/null`?

palladia:
No difference on the order of 2>&1 and >/dev/null

Klaas-:
When is this hitting azure production? A rhel VM I rolled out last week still has:
```
# crontab -l
* * * * * [ \( ! -f /etc/opt/omi/creds/omi.keytab \) -o \( /etc/krb5.keytab -nt /etc/opt/omi/creds/omi.keytab \) ] && /opt/omi/bin/support/ktstrip /etc/krb5.keytab /etc/opt/omi/creds/omi.keytab
```

Also I think your check is insufficient to update a preexisting configuration
https://github.com/Microsoft/omi/blob/master/Unix/scripts/config_keytab_update.sh#L15
is true for the "old" crontab line and so it won't replace it with a new one if config_keytab_update.sh is run. Not sure if that is actually the workflow you use though :)

Klaas-:
I think the issue is that rhel is not shipping the microsoft production repo. Updating the omi rpm actually removes the crontab and then readds it, so that is fine for me.

JumpingYang001:
@Klaas-, thanks for adding comments here, we are providing a new option in omi 1.6.0 `sudo /opt/omi/bin/support/config_keytab_update.sh  --purgeforever` to remove the Kerberos crontab forever, if you don't need the Kerberos crontab it is a new option.

For MS repo, we will update it once omi 1.6.0 release, thanks.

PR #517, opened 2018-05-07
Let installbuilder handle sockets directory

palladia:


JumpingYang001:
I  still think this doesn't fix the real root cause issue, since install omi standalone will not have issue, the issue is very possible caused by omsconfig or dsc installer or oms plugins or provider, but will not be omi itself.

JumpingYang001:
BTW, I see other installer is set permission for nonroot user in postinstall like https://github.com/Microsoft/OMS-Agent-for-Linux/blob/4393ab453bbd4bfc3c7e0995f0b759a61fbbce44/installer/datafiles/base_omsagent.data#L166

JumpingYang001:
I still don't find where set sockets folder to omsagent:root.

PR #524, opened 2018-06-01
Fixes double free issue

JumpingYang001:
@Microsoft/omi-devs @dantraMSFT  , fixed the double free issue #502 , could you help to review it? thanks.

### DIFF
--- a/Unix/miapi/InteractionProtocolHandler.c
+++ b/Unix/miapi/InteractionProtocolHandler.c
@@ -557,8 +557,16 @@ static void InteractionProtocolHandler_Operation_Strand_Finish( _In_ Strand* sel
         Message_Release(&operation->req->base);
     }
     SessionCloseCompletion_Release(operation->sessionCloseCompletion);
-    PAL_Free(operation->protocolConnection);
-    PAL_Free(operation);
+    if(operation->protocolConnection)
+    {
+        PAL_Free(operation->protocolConnection);
+        operation->protocolConnection = NULL;
+    }
+    if(operation)
+    {
+        PAL_Free(operation);
+        operation = NULL;
+    }
 }
 
 #ifdef _PREFAST_
@@ -633,7 +641,11 @@ MI_Result MI_CALL InteractionProtocolHandler_Operation_Close(
     else
     {
         SessionCloseCompletion_Release(operation->sessionCloseCompletion);
-        PAL_Free(operation);
+        if(operation)
+        {
+            PAL_Free(operation);
+            operation = NULL;
+        }
     }
 
     return MI_RESULT_OK;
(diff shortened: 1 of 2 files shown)


JumpingYang001:
@palladia you can check other place for "PAL_Free(operation->protocolConnection);" in interationprotocolhandler.c

pal_free will call systemfree which will call free which will not set it to null, Setting unused pointers to NULL is a defensive style, protecting against dangling pointer bugs.  previous logic operation->connection become unused pointer here, then when next checking in other place it is not null and can be free, then double free occurs.

next line operation also need to set to null, I will update it if no other issues, thanks.

palladia:
I think you misunderstood me.  I do know what PAL_Free() does, and I do know why one sets the pointer to NULL after free.  I'm just not sure this will fix Dan's issue.  But let's wait for Dan's verification.

Anyway, if you want to be defensive, do something like:

if (operation)
{
    PAL_Free(operation->protocolConnection);
    operation->protocolConnection = NULL;
    PAL_Free(operation);
    operation = NULL;
}

dantraMSFT:
I don't see how this can fix https://github.com/Microsoft/omi/issues/502 either.

The problem I saw was as follows:

InteractionProtocolHandler_Session_Connect calls WsmanClient_New_Connector with an allocated operation.  Before returning, the connection fails and the operation is freed.   This would be fine if WsmanClient_New_Connector returned a failing MI_RESULT  but it doesn't, it returns MI_RESULT_OK and the code assumes the operation is still a valid memory pointer.  As a result, the following line is executed:

if (operation->protocolConnection->protocol.wsman == NULL)

Since the operation has been freed, in debug builds, protocolConnection contains an invalid pointer value and we get a seg fault.  In release builds, we're accessing freed memory and it's state is indeterminate.

Even with the suggested changes, we'll get a seg fault because we'll be dereferencing a null pointer, protocolConnection.   If we add a protocolConnection null pointer check, we'll end up getting a random seg fault any time that memory block happens to get reused for something else.

The reliable fix needs to handle the case that operation itself has been freed.

JumpingYang001:
thanks all! I got understand now. but after set to null, the double free Looks like disappear. anyway I will update the PR to make it sense.

JumpingYang001:
@paulcallen @dantraMSFT , can you help to review the new changes? thank you!

JumpingYang001:
@paulcallen @dantraMSFT , ping...

JumpingYang001:
@paulcallen @dantraMSFT , ping...

JumpingYang001:
@paulcallen , thanks for confirm! I'll debug the failed Pester test with double free issue to understand the underlying problem.

JumpingYang001:
I don't have a good fix at present, so I want to close this PR, I will new a PR once I have a good fix, thanks.

PR #527, opened 2018-06-07
Set omi:omi as omi.keytab's owner

BingyuLiu:


JumpingYang001:
Negotiate file also need to chown : `[ -d /etc/opt/omi/.creds] && chown -R omi:omi /etc/opt/omi/.creds`
```
/etc/opt/omi/.creds/
/etc/opt/omi/.creds/ntlm
```

JumpingYang001:
another file is `/etc/opt/omi/ssl/omikey.pem`, let me check today if it will be deleted after uninstallation.

JumpingYang001:
I have confirmed that `/etc/opt/omi/ssl/omikey.pem` will exist after uninstallation on Negotiate environment.

BingyuLiu:
@JumpingYang001 thanks, please review it again!

PR #528, opened 2018-06-08
Add sslciphersuite usage in readme

JumpingYang001:
one customer report they want to disable some weak ciphers, I want to document these settings, @Microsoft/omi-devs , could you help to review it? thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ NoSSLv3            | When `true`, the SSLv3 protocol is disabled. If NoSSLv2 and
 NoTLSv1_0          | When `true`, the TLSv1.0 protocol is disabled
 NoTLSv1_1          | When `true`, and if available on the platform, the TLSv1.1 protocol is disabled
 NoTLSv1_2          | When `true`, and if available on the platform, the TLSv1.2 protocol is disabled
-sslCipherSuite     | The prioritized list of allowed SSL/TLS ciphers. For more information, check `man ciphers` or search internet with `openssl man ciphers`
+sslciphersuite     | The prioritized list of allowed SSL/TLS ciphers. For example, set `sslciphersuite=ALL:!SSLv2:!SSLv3:!TLSv1:!RC4-MD5:!RC4-SHA:!SEED-SHA` in `/etc/opt/omi/conf/omiserver.conf` to disable all SSLv2,SSLv3,TLSv1 ciphers and 3 weak ciphers: RC4-MD5,RC4-SHA,SEED-SHA; then run `sudo /opt/omi/bin/service_control restart` to take effect, for more information, check `man ciphers` or search internet with `openssl man ciphers` 
 
 ### Configuring OMI Client
 


README.md
```diff
@@ -140,7 +140,7 @@ NoSSLv3            | When `true`, the SSLv3 protocol is disabled. If NoSSLv2 and
 NoTLSv1_0          | When `true`, the TLSv1.0 protocol is disabled
 NoTLSv1_1          | When `true`, and if available on the platform, the TLSv1.1 protocol is disabled
 NoTLSv1_2          | When `true`, and if available on the platform, the TLSv1.2 protocol is disabled
-sslCipherSuite     | The prioritized list of allowed SSL/TLS ciphers. For more information, check `man ciphers` or search internet with `openssl man ciphers`
+sslCipherSuite     | The prioritized list of allowed SSL/TLS ciphers. For example, run `sudo echo 'sslciphersuite=ALL:!SSLv2:!SSLv3:!TLSv1:!RC4-MD5:!RC4-SHA:!SEED-SHA'>> /etc/opt/omi/conf/omiserver.conf; sudo /opt/omi/bin/service_control restart` to disable all SSLv2,SSLv3,TLSv1 ciphers and 3 weak ciphers: RC4-MD5,RC4-SHA,SEED-SHA(note:if omiserver.conf contains sslciphersuite before run the command, please modify the file to comment it or remove it.), for more information, check `man ciphers` or search internet with `openssl man ciphers`
```
palladia: 1.  sslciphersuite in omiserver.conf seems to be case sensitive, so either fix the code in servercommon.c, or modify the README to lowercase only.
2.  I don't like asking users to append to omiserver.conf, as it's easy to make mistakes, or have duplicate lines.  It's easier to just ask them to edit the conf file.
3.  I am a bit confused about ciphers vs. cipher suites.  Options like "noTLSv1_1" set the cipher suite via SSL_CTX_set_options()?  But ciphers like "RC4-MD5" are set via SSL_CTX_set_cipher_list()?   Probably can't change the conf option name "sslCipherSuite", but it should only contain ciphers, so no "!SSLv2:SSLv3:!TLSv1"?

JumpingYang001: I will update it, thanks.

JumpingYang001: for number 3, doc shows `It can represent a list of cipher suites containing a certain algorithm, or cipher suites of a certain type. For example SHA1 represents all ciphers suites using the digest algorithm SHA1 and SSLv3 represents all SSL v3 algorithms.` so my usage should be correct.
https://www.openssl.org/docs/man1.0.2/apps/ciphers.html

JumpingYang001:
@palladia , I have resolved the commented issues, please help to review again if I need to update again, thank you!

for the 3rd point, official doc shows `It can represent a list of cipher suites containing a certain algorithm, or cipher suites of a certain type. For example SHA1 represents all ciphers suites using the digest algorithm SHA1 and SSLv3 represents all SSL v3 algorithms.` so my usage should be correct.
https://www.openssl.org/docs/man1.0.2/apps/ciphers.html

JumpingYang001:
thanks  for reviewing! We have this kinds of testing at  the beginning of OMI project testing, but it isn't in current test suites for some history reason, we'll add them back.

JumpingYang001:
doc say both ciphersuite(SHA1) and ciphersuite type(TLSv1) are fine.

PR #530, opened 2018-06-16
Fix Heimadal Kerberos fail in ktstrip

### LINKED ISSUE #529: ktstrip cron job not calling ktutil correctly
Running Ubuntu 16.04 in Azure with Heimdal Kerberos.  

Every minute, we're receiving the following email for around 80 systems:
Subject: 
> Cron <root@ase1cp-itmg-001> [ -f /etc/krb5.keytab ] && [ \( ! -f /etc/opt/omi/creds/omi.keytab \) -o \( /etc/krb5.keytab -nt /etc/opt/omi/creds/omi.keytab \) ] && /opt/omi/bin/support/ktstrip /etc/krb5.keytab /etc/opt/omi/creds/omi.keytab || true

Body:
> src kt = /etc/krb5.keytab
> Usage: ktutil [-hv] [--version] [--help] [--keytab=keytab to operate on]
>    [-k keytab to operate on] [--verbose] command
> -k keytab to operate on, --keytab=keytab to operate on keytab
> -v, --verbose                                          verbose
> Creating /etc/opt/omi/creds/omi.keytab
> Usage: ktutil [-hv] [--version] [--help] [--keytab=keytab to operate on]
>    [-k keytab to operate on] [--verbose] command
> -k keytab to operate on, --keytab=keytab to operate on keytab
> -v, --verbose                                          verbose
> Move /tmp/tmp.kt to /etc/opt/omi/creds/omi.keytab
> mv: cannot stat '/tmp/tmp.kt': No such file or directory
> chown: cannot access '/etc/opt/omi/creds/omi.keytab': No such file or directory
> could not set user   omi:omi  as owner of /etc/opt/omi/creds/omi.keytab
> 

I think the assumption by the /opt/omi/bin/support/ktstrip script that the system is running MIT Kerberos is possibly to blame here.  My reasoning follows.

Reading through /opt/omi/bin/support/ktstrip, it appears that it's trying to do the equivalent of the following command:

`echo -e '/etc/krb5.keytab\nlist\n' |  ktutil`

which implies that ktutil has an interactive mode, but it doesn't. It only accepts command line arguments.  The correct command should be:

`ktutil -k /etc/krb5.keytab list`

It's possible that not only the command format is different, but the output is different as well, but I didn't install MIT Kerberos to determine if that's the case or not.

JumpingYang001:
@Microsoft/omi-devs @palladia  @twstdpear  , fixes #529 , could you help to review it? thanks.

### DIFF
--- a/Unix/tools/ktstrip
+++ b/Unix/tools/ktstrip
@@ -59,7 +59,11 @@ if [ $? -ne 0 ]; then
    echo "ktutil does not exist" 2>&1
    exit 3
 fi    
-printf "read_kt $SRC_KT\nlist\n" | ktutil >$list  
+printf "read_kt $SRC_KT\nlist\n" | ktutil >$list 2>&1 >/dev/null
+if [ $? -ne 0 ]; then
+   echo "we only support MIT Kerberos on Linux at present" 2>&1
+   exit 3
+fi
 
 #
 # Take the list. The first awk selects only the lines that start with some nomber of spaces and a digit. This gest rid of the 


JumpingYang001:
the best way is to implement Heimadal.

JumpingYang001:
I don't have other better idea about the condition to distinguish them.

PR #534, opened 2018-07-02
Support MS-WSMAN session cookies over HTTP

dantraMSFT:
This change adds support for detecting and using the MS_WSMAN session cookie.  This is needed to maintain connections to sites behind a load balancer, such as office 365.

The change is in the following layers, 
* httpclient (httpclient.c, httpclient_private.h, httpclientauth.c), 
* wsmanclient (wsmanclient.c)
* Interaction protocol (InteractionProtocolHandler.c)
* Message.h (updates the base Message structure to support a session cookie)

The logic starts at the response from the target containing a Set-Cookie: MS-WSMAN:* header.  This is handled by the wsmanclient layer which detects and passes up using the base Message structure (typically via PostInstance)

InteractionProtocolHandler.c is responsible for detecting changes to the cookie value and saving it for the next request sent to the server. At that point, the session cookie is added to the MI_DestinationOptions passed down to create the connection.  Note that not all messages sent from wsmanclient to interaction protocol contains a session cookie.

Finally, the http layer is responsible for adding the Cookie header on the next request.

### DIFF
--- a/Unix/base/messages.h
+++ b/Unix/base/messages.h
@@ -195,6 +195,13 @@ struct _Message
      */
     MI_Char *shellId;
 
+    /* Message's destructor [opt]
+        'Release' will call dtor (if set) right before destroying the message */
+    void (*dtor)(Message* message, void* callbackData);
+
+    /* Data passed as 2nd argument of 'dtor' */
+    void* dtorData;
+
     /*
         For http transport operations that go through a load balancer,
         we need the MS_WSMAN cookie for subsequent requests.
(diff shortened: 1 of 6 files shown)


Unix/base/messages.h
```diff
@@ -195,6 +195,14 @@ struct _Message
...
     MI_Char *shellId;
 
+    /*
+        For http transport operations that go through a load balancer,
+        we need the MS_WSMAN cookie for subsequent requests.
+        The value is retrieved from the HttpHeaders passed to
+        _HttpProcessRequest.
+    */
+    MI_Char *sessionCookie;
+
```
palladia: Could you move this new member to end of structure?  We've seen this before, where adding an item in middle of struct causes backward incompatibility issues.

dantraMSFT: @palladia: I made the change but I don't see how it solves anything. Message is referenced inline by all other messages so there will be size and alignment changes regardless.  Is there some other backwards compatibility issue you're thinking about?

dantraMSFT:
Moving this to a branch in the OMI repo.

PR #539, opened 2018-07-06
Support MS-WSMAN session cookies over HTTP

dantraMSFT:
This change adds support for detecting and using the MS_WSMAN session cookie. This is needed to maintain connections to sites behind a load balancer, such as office 365.

The change is in the following layers,

httpclient (httpclient.c, httpclient_private.h, httpclientauth.c),
wsmanclient (wsmanclient.c)
Interaction protocol (InteractionProtocolHandler.c)
Message.h (updates the base Message structure to support a session cookie)
The logic starts at the response from the target containing a Set-Cookie: MS-WSMAN:* header. This is handled by the wsmanclient layer which detects and passes up using the base Message structure (typically via PostInstance)

InteractionProtocolHandler.c is responsible for detecting changes to the cookie value and saving it for the next request sent to the server. At that point, the session cookie is added to the MI_DestinationOptions passed down to create the connection. Note that not all messages sent from wsmanclient to interaction protocol contains a session cookie.

Finally, the http layer is responsible for adding the Cookie header on the next request.

### DIFF
--- a/Unix/base/messages.h
+++ b/Unix/base/messages.h
@@ -201,6 +201,14 @@ struct _Message
 
     /* Data passed as 2nd argument of 'dtor' */
     void* dtorData;
+
+    /*
+        For http transport operations that go through a load balancer,
+        we need the MS_WSMAN cookie for subsequent requests.
+        The value is retrieved from the HttpHeaders passed to
+        _HttpProcessRequest.
+    */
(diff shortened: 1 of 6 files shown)


JumpingYang001:
it would be better to add new unittests since the code added new feature.

Issue #523, opened 2018-05-30
Handle session cookies over HTTP to reliably work with office365.


dantraMSFT:
PowerShell has received a number of reports of errors when connecting to office365 through https://outlook.office365.com/powershell-liveid/.

The symptoms are as follows:
1: Connect to office 365
$Session = New-PSSession -ConfigurationName Microsoft.Exchange -ConnectionUri https://outlook.office365.com/powershell-liveid/ -Credential $O365Credential -Authentication Basic -AllowRedirection

2: Perform various commands

3: At somewhat random times, commands on the connection start returning ERROR_WSMAN_INVALID_SELECTORS and the connection is no longer usable.

Analysis of the logs shows office365 is behind a load balancer and returns a cookie in the response.
Set-Cookie: MS-WSMAN=S-1-5-21-2553855760-3398840149-2091067650-9554679=rJqNiZqNgbKot6+tzc6yvc/Nx8fRkZ6Sj42bzc7Rj42Qm9GQiouTkJCU0ZyQkoHOxsvMzM/Iz8fOgc3PzsfSz8rSzcarzsjFzsrFysg=; expires=Tue, 29-May-2018 17:15:57 GMT; path=/powershell-liveid; secure; HttpOnly

The problem can be reproduced but no consistently since it's subject to load balancing.

The PowerShell issue tracking the problem is https://github.com/PowerShell/PowerShell/issues/3771

JumpingYang001:
This looks like a DCR, @Microsoft/omi-devs

paulcallen:
it kind of is a DCR. However it was believed that psrp remoting worked against office365 and this is blocking that from working.
If a cookie is returned for a request on the session it should be included in all future requests on that session.

dantraMSFT:
@JumpingYang001 PSRP does work with office365 until the load balancer directs requests to another server.  I've seen it remain valid for multiple hours and also seen it fail in ~1 minute.  At that point, the session cannot be used.

JumpingYang001:
I didn't use psrp/PowerShell to remote to office365, Let me try it first and then try to understand the problem, thanks.

paulcallen:
The problem is understood! The http client code needs to support cookies!

JumpingYang001:
@paulallen could you help to explain the logic? is my below understand right or wrong?
1. powershell call psrp, then psrp call libmi.so to send http request to office365.
2. libmi.so get http response from office365, and then send it back to psrp, then powershell get the data.

[...]

PR #543, opened 2018-07-16
Handle session cookies to office365.com

dantraMSFT:
Fix https://github.com/Microsoft/omi/issues/523
This change places management of the session cookie in the HttpClient layer.  There are four basic parts:

* TransactionProtocolHandler: Generates a session id for the current MI_Session and passes it to the HttpClient layer via MI_DestinationOptions.
* SessionMap - a HashMap providing a mapping between a session id and an HTTP cookie.
* HttpClient.c - adds/updates the SessionMap when a response is recieved with a MS-WSMAN cookie using the id created by TransactionProtocolHandler.  Ensures subsequent requests also include the session cookie.
NOTE: With office365, the session cookie changes virtually on every response and a single cookie cannot be used.
* TransactionProtocolHandler removes the session id from the SessionMap when the session is destroyed. This is done via a call to a new  export HttpClient_RemoveSession.

NOTE: The current implementation initializes the HashMap on first use since there is no clear global initialization point.  Since there is also no clear global cleanup, the internal HashMap array is also leaked on shutdown.  Allocations for hashMap entries should not be leaked since MI_Session clean removes the associated entry.

Unlike the previous attempt, this approach is able to track the session cookie for all types of responses and solves the case where an error, such as an timeout, can cause the cookie to stale for the next request.

### DIFF
--- a/Unix/http/sessionmap.c
+++ b/Unix/http/sessionmap.c
@@ -1,0 +1,323 @@
+#include "sessionmap.h"
+#include <pal/memory.h>
+#include <pal/strings.h>
+#include <common.h>
+
+#define ENABLE_TRACING 1
+#define FORCE_TRACING 1
+
+#ifdef ENABLE_TRACING
+# define TRACING_LEVEL 4
+#include <base/result.h>
+#include <base/logbase.h>
+#include <base/log.h>
+#define LOGE2 __LOGE
+#define LOGW2 __LOGW
+#define LOGD2 __LOGD
+#else
+# define LOGE2(a)
+# define LOGW2(a)
+# define LOGD2(a)
+# define LOGX2(a)
+#endif
+
+typedef struct _SessionBucket
+{
+    struct _SessionBucket *next;// HashMap usage
+    const char* sessionId;      // WS_MAN Session Id 
+    size_t hash;                // Hash value for sessionId
+    char* cookie;               // MS-WSMAN Cookie
+                                // allocated on creation and update.
+                                // owned by the bucket
+} SessionBucket;
+
+#define SESSIONID_HASH_SEED 1313038763
+
+static size_t _SessionId_Hash(_In_z_ const char* sessionId)
+{
+    size_t hash = SESSIONID_HASH_SEED;
+    if (sessionId)
+    {
+        const char*p = sessionId;
+        while(*p)
+        {
+            // From strset.c
+            hash ^= ((hash << 5) + *p + (hash >> 2));
+            p++;
+        }
+    }
+    return hash;
+}
+
+static size_t _SessionBucket_GetHash(_In_ const HashBucket *b)
+{
+    const SessionBucket *bucket = (const SessionBucket*) b;
+    return bucket->hash;
+}
+
+static int _SessionBucket_Equal(_In_ const HashBucket* b1, _In_ const HashBucket* b2)
+{
+    const SessionBucket* bucket1 = (const SessionBucket*) b1;
+    const SessionBucket* bucket2 = (const SessionBucket*) b2;
+    return 
+    (
+        bucket1->hash == bucket2->hash
+        &&
+        strcmp(bucket1->sessionId, bucket2->sessionId) == 0
+    );
+}
+
+static void _SessionBucket_Release(HashBucket *b)
+{
+    if (b)
+    {
+        SessionBucket* bucket = (SessionBucket*) b;
+        if (bucket->cookie)
+        {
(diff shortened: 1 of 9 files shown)


Unix/http/sessionmap.c
```diff
@@ -0,0 +1,318 @@
...
+    SessionBucket* bucket = NULL;
+    if (sessionId) do
+    {
+        size_t keyLen = strlen(sessionId);
+        size_t size = sizeof(SessionBucket) + keyLen + 1;
+        bucket = PAL_Malloc(size);
+        if (NULL == bucket)
+        {
+            LOGE2((ZT("_SessionBucket_New - failed to allocate bucket: %s"), sessionId));
+            DEBUG_ASSERT(FALSE);
```
paulcallen: out of memory should not assert

Unix/http/sessionmap.c
```diff
@@ -0,0 +1,318 @@
...
+            break;
```
paulcallen: you are going to crash in a few lines. You should just return NULL.

Unix/http/sessionmap.c
```diff
@@ -0,0 +1,318 @@
...
+        }
+        PAL_Free(bucket);
+    }
+}
+
+static SessionBucket* _SessionBucket_New(SessionMap *self, _In_z_ const char* sessionId, size_t hash)
+{
+    // WriteLock expected.
+    SessionBucket* bucket = NULL;
+    if (sessionId) do
```
paulcallen: Personally I would prefer if and do on different lines, where do is nested inside curly brackets. It is cleaner to read.

Unix/http/sessionmap.c
```diff
@@ -0,0 +1,318 @@
...
+        bucket->cookie = NULL;
+        bucket->next = NULL;
+
+        if (HashMap_Insert (&self->map, (HashBucket*) bucket) != 0)
+        {
+            // Not expecting the bucket to already exist.
+            _SessionBucket_Release((HashBucket*) bucket);
+            bucket = NULL;
+            LOGE2((ZT("_SessionBucket_New - sessionId already exists: %s"), sessionId));
+            DEBUG_ASSERT(FALSE);
```
paulcallen: don't assert, just fail gracefully

Unix/http/sessionmap.c
```diff
@@ -0,0 +1,318 @@
...
+        }
+
+        char* clone = (char*) (((unsigned char*) bucket) + sizeof(SessionBucket));
+        strcpy(clone, sessionId);
+        bucket->sessionId = clone;
+        bucket->hash = hash;
+        bucket->cookie = NULL;
+        bucket->next = NULL;
+
+        if (HashMap_Insert (&self->map, (HashBucket*) bucket) != 0)
```
paulcallen: map is not protected with a lock

Unix/http/sessionmap.c
```diff
@@ -0,0 +1,318 @@
...
+
+        if (0 == result)
+        {
+            LOGD2((ZT("_SessionMap_IsValid - SessionMap initialized.")));
+            self->state = SESSIONID_HASH_SEED;
+        }
+        else
+        {
+            LOGE2((ZT("_SessionMap_IsValid - Could not initialize the HashMap.")));
+            DEBUG_ASSERT(FALSE);
```
paulcallen: asserting on failure is wrong. Return an error and logging is sufficient.

Unix/http/sessionmap.c
dantraMSFT: Where do you see the crash?  The break statement breaks out of the while look and simply returns bucket; which is NULL due to the failed allocation.

Unix/http/sessionmap.c
dantraMSFT: current style cop tools seem to recommend this pattern. Personally, I prefer it as well to avoid the extra indenting.    I'll add a // while(FALSE) comment for clarity unless you feel strongly enough about it.

Unix/http/sessionmap.c
dantraMSFT: I'll add a comment to the function. the caller must acquire a write lock and it does because it must first search.  See SessionMap_SetCookie

Unix/http/sessionmap.c
dantraMSFT: No fatal errors, right. :)

paulcallen:
can you squash your change list. Don't need intermediate changes in the history. thanks.

Issue #544, opened 2018-07-17
httpClient: Log a timestamp in omiclient-send.trc to support correlating with omiclient-recv.trc


dantraMSFT:
When debug logging of http traffic, responses include a GMT date stamp in the header but the requests do not. This makes it difficult to correlate requests with responses.

The fix would be to log a GMT timestamp in the block where the the client headers are logged; something like the following:

if (FORCE_TRACING || handler->enableTracing)
{
     // log  a timestamp before the client headers
     char timestamp[40];
     GetUtcTimeStamp(timestamp, 40);
     _WriteTraceFile(ID_HTTPCLIENTSENDTRACEFILE, timestamp, Strlen(timestamp);

    _WriteTraceFile(ID_HTTPCLIENTSENDTRACEFILE, buf, buf_size);
}

JumpingYang001:
looks like it doesn't recv file or send file special, I found timestamp here: https://github.com/Microsoft/omi/blob/master/Unix/base/log.c#L204

dantraMSFT:
@JumpingYang001: I'm not sure what you're referring to.

The current httpclient code doesn't log with timestamps for http debug tracing.  It's not an issue for responses since it contains a Date header.  Requests do not.

This is what I've been using to debug session cookie issues.
https://github.com/Microsoft/omi/commit/7fdde2eec4f9aefeeb8210c52dcd3bc090417229

It probably makes sense to also log the generated session id to improve correlation.

JumpingYang001:
ok, looks like my provided link is about normal logging, and trace logging is in another function: https://github.com/Microsoft/omi/blob/ba1958ec946fc7705550d4d78796d5719e54c2cd/Unix/http/http.c#L398.  BTW, I don't see any log timestamp in response trace log code.  will check it today.

dantraMSFT:
The response headers have a Date header so I didn't need to log one.  It makes sense though to log a session id/timestamp logged for each header set.

PR #545, opened 2018-07-18
Log a session id and datetime stamp at the start of logging HTTP headers

dantraMSFT:
Fix: https://github.com/Microsoft/omi/issues/544

The intent of this change is to provide a session id and date/time stamp in the debug logs for httpclient (omiclient-sent.trc and omiclient-recv.trc).  It aids in correlating http traces with a given session id and includes the timestamp for the sent/recieved http headers.

Previously, omiclient-sent.trc contained no date/time information and omiclient-recv.trc relied on the Date header in the response.  This made correlating between the two log files error prone, at best.

* Example omiclient-recv.trc log:
[Session: 1 Date: 2018-07-18 00:31:36.0449595Z]
HTTP/1.1 200 OK
Cache-Control: private
Content-Type: application/soap+xml;charset=UTF-8
...

* Example omiclient-recv.trc log:
[Session: 1 Date: 2018-07-18 00:31:17.0553452Z]
POST /powershell-liveid?PSVersion=6.1.0 HTTP/1.1
Connection: Keep-Alive
Content-Length: 47119
Content-Type: application/soap+xml;charset=UTF-8
...


Unix/http/httpclient.c
```diff
@@ -540,6 +541,48 @@ static MI_Result _Sock_Write(
...
+
+    NOTES: This is called by _WriteClientHeader using ID_HTTPCLIENTRECVTRACEFILE
+    and _ReadHeader using ID_HTTPCLIENTSENDTRACEFILE.
+    It provides a marker to indicate the start of a request or response in the omiclient-*.trc 
+    file as well as aiding in correlating messages to a given session.
+*/
+static void _WriteTraceSessionTimestamp(_In_ PathID pathId, _In_opt_z_ const char* sessionId)
+{
+    // session id: 20, timestamp: 29, whitespace and static text:19, zero terminator
+    // [Session: NNNNN Date: YYYY/MM/DD HH:MM:SS.1234567Z]
```
JumpingYang001: `YYYY/MM/DD` and `%04d-%02d-%02d` need to use same delimiter, thanks.

dantraMSFT: @JumpingYang001 - I chose the format because it encodes UTC and milliseconds. While the milliseconds are not that important, the UTC is because it can be correlated to the Date header included in the response which is also UTC.

JumpingYang001: I mean comment used `/` but the real log is dash`-`
possible better comment: `// [Session: NNNNN Date: YYYY-MM-DD HH:MM:SS.1234567Z]`

PR #555, opened 2018-07-21
Fixes preexec issue with multiple requests

palladia:
The class _ConnectionIn uses operationId as key to check for duplicate entries in hash table.  This causes multiple preexec requests to be ignored.


JumpingYang001:
Verified the fixes on CentOS6x64 and Ubuntu16.01x64

PR #565, opened 2018-08-24
Fixes thunk handle

### LINKED ISSUE #448: ERROR: "Trying to thunk generic handle that has been shutdown already"
omi build: `\\redmond\wsscfs\OSTCData\Builds\omi\release\1.4.1-0`

**NOTE:** 1.3.0-2 also has this issue, but 1.2.0-35 don't have miclient.log.

**Repro Systems:**
Ubuntu 16.04 x64, Suse 12x64, Debian 9x64.
**NOTE:** No issue systems: Redhat7x64,CentOS7x64,Oracle7x64.

**Reproduce steps:**
1. Install omi
2. Run `while true; do /opt/omi/bin/omicli id; done`
3. Check `cat /var/opt/omi/log/miclient.log|grep ERROR` and you will see one new error "2017/11/19 23:48:04 [24799,24799] ERROR: null(0): EventId=20106 Priority=ERROR Trying to thunk generic handle that has been shutdown already: 0x7ffc9c3909f0" per new query.

Actual result: ERROR in miclient.log
Expect result: No error in miclient.log

JumpingYang001:
Fixed #448 , @palladia @paulcallen , can you help to review it? thanks.

### DIFF
--- a/Unix/base/oi_traces.h
+++ b/Unix/base/oi_traces.h
@@ -333,8 +333,6 @@ OI_EVENT("Trying to thunk invalid generic handle: %p")
 void trace_MIThunk_InvalidHandle(void * handle);
 OI_EVENT("Trying to thunk generic handle that has an old version number: %p")
 void trace_MIThunk_OldVersion(void * handle);
-OI_EVENT("Trying to thunk generic handle that has been shutdown already: %p")
-void trace_MIThunk_AlreadyShutdown(void * handle);
 
 OI_EVENT("Timer_Start:Timer (%p) already running")
 void trace_Timer_CannotStartTimer_AlreadyRunning(void* timer);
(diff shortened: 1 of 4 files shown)


paulcallen:
cancelling an already shutdown handle is OK as long as the application is not shut down which implies more serious problems in the flow of an application not this code. 
What you have done here is nothing more than an unreliable band aid that really does not solve the threading issue. If you don't like the error, make it a warning! This error is likely to happen in ordinary code so it is an interesting warning of misuse of the handle, but not really a problem. These are thunk handles to protect race conditions like the one you are seeing. It makes doing what is happening here safe. If it is crashing that is a different story, but I believe you just have a problem with the debug message and find it alarming.

paulcallen:
but it doesn't really solve the underlying problem and can still happen due to multiple threads. I would just change the logging if it worries anyone!

JumpingYang001:
thanks Paul! I want change it to INFO since you said cancelling an already shutdown handle is OK , will update the changes.

PR #566, opened 2018-08-24
-h option is not recognized in new openssl in ubuntu18

BingyuLiu:


Unix/configure
```diff
@@ -2243,8 +2243,11 @@ fi
 openssl -h > /dev/null 2> /dev/null
 
 if [ "\$?" != "0" ]; then
-    echo "\$0: openssl program is not executable."
-    exit 1
+    openssl help > /dev/null 2> /dev/null
+    if [ "\$?" != "0" ]; then
+        echo "\$0: openssl program is not executable."
+        exit 1
+    fi
```
palladia: openssl -h doesn't work on ubuntu 16.04 either, but it returns a '0' return code.

Instead of running "openssl -h", which doesn't seem to be a supported command, why not just run "openssl version", which (should) works on all platforms?

BingyuLiu: I don't know, have to try.

BingyuLiu: 'openssl version' works fine in all platform. thanks!

PR #571, opened 2018-09-05
Fixes leak temp files

### LINKED ISSUE #570: OMI cron with Kerberos creating boot issues - 1 hour to reboot
I have this issue on many of my Azure VMs: rebooting them takes a really long time, an hour or more.
What I have found is that the OMI cron fails to remove the temporary files and this makes my /tmp directory to have up to 1 million files in it, and adding 3 new files every minute:

```
tmp.001gqDWeTO
tmp.003kaLPkdz
tmp.005lyx3CFp
```

The cron exit code is 0, but a fake 0 because of that "OR true" option:

```
# [ -f /etc/krb5.keytab ] && [ \( ! -f /etc/opt/omi/creds/omi.keytab \) -o \( /etc/krb5.keytab -nt /etc/opt/omi/creds/omi.keytab \) ] && /opt/omi/bin/support/ktstrip /etc/krb5.keytab /etc/opt/omi/creds/omi.keytab >/dev/null 2>&1 || true

# echo $?
0
```

However this part of the cron command errors out:

```
# /opt/omi/bin/support/ktstrip /etc/krb5.keytab /etc/opt/omi/creds/omi.keytab
src kt = /etc/krb5.keytab
Creating /etc/opt/omi/creds/omi.keytab
ktutil:  read_kt /etc/krb5.keytab 
ktutil:  delete_entry 1
ktutil:  write_kt /tmp/tmp.kt
ktutil:  Move /tmp/tmp.kt to /etc/opt/omi/creds/omi.keytab
mv: cannot stat '/tmp/tmp.kt': No such file or directory
chown: cannot access '/etc/opt/omi/creds/omi.keytab': No such file or directory
could not set user   omi:omi  as owner of /etc/opt/omi/creds/omi.keytab
```

Having so many files in /tmp is a problem at reboot time. It takes a really long time to remove them all.

Since this cron job is somehow useless I've tried to remove it, but something adds it back after reboot. Could be the OMI service restart.

My VM has the /etc/krb5.keytab file. SSH login is managed by central AD and some Windows AD groups per sshd configuration.

What is the intended purpose of this cron? Since it never worked I probably do not miss it. But I would like to know.
Can this cron be either removed or at least fixed so it will not leave behind all those tmp files? #425

JumpingYang001:
Fixed #570 , @Microsoft/omi-devs , could you help to review it? thank you!
Fixed result:
```
root@ost64-ct7-06.scx.com  # /opt/omi/bin/support/ktstrip /tmp/newtmp.kt /etc/opt/omi/creds/omi.keytab
src kt = /tmp/newtmp.kt
Creating /etc/opt/omi/creds/omi.keytab
ktutil:  read_kt /tmp/newtmp.kt 
ktutil:  delete_entry 1
ktutil:  write_kt /tmp/tmp.kt
ktutil:  Cleaning temp files...
File  /tmp/tmp.kt  is not generated, so we skip configure Kerberos authentication for omi.

~
root@ost64-ct7-06.scx.com  # echo $?
0 
```

### DIFF
--- a/Unix/scripts/config_keytab_update.sh
+++ b/Unix/scripts/config_keytab_update.sh
@@ -19,6 +19,11 @@ checkConfiguration()
     return 0
 }
 
+cleanTempFile()
+{
+    # clean temp file
+    rm -f $tmpfile > /dev/null 2>&1
+}
 
 configure()
 {
@@ -58,6 +63,9 @@ unconfigure()
     grep -v "$omikeytab" $tmpfile >$tmpfile2 
     crontab $tmpfile2
 
+    # clean temp file
+    rm -f $tmpfile2 > /dev/null 2>&1
+
     timestamp=$(date +%F\ %T)
     printf "$timestamp : Crontab no longer configured to update omi keytab.\n"
 }
@@ -84,3 +92,4 @@ do
     esac
 done
 
+cleanTempFile
--- a/Unix/tools/ktstrip
+++ b/Unix/tools/ktstrip
@@ -49,6 +49,15 @@ delete_list=$(mktemp)
 cmd=$(mktemp)
 tmp_kt="/tmp/tmp.kt"
 
+# clean temp files
+CleanTempFiles()
+{
+    echo "Cleaning temp files..." 2>&1
+    rm -f $list > /dev/null 2>&1
+    rm -f $delete_list > /dev/null 2>&1
+    rm -f $cmd > /dev/null 2>&1
+}
+
 #
 # Use ktutil to strip the soure keytab of anything but host/ and RestrictedKrbHost/ principals
 # First, get a list of principals and put it in a temp file to retain its line structure. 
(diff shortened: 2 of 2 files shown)


Unix/scripts/config_keytab_update.sh
```diff
@@ -58,6 +63,10 @@ unconfigure()
     grep -v "$omikeytab" $tmpfile >$tmpfile2 
     crontab $tmpfile2
 
+    # clean temp file
+    cleanTempFile
```
palladia: You don't need to call cleanTempFile here, since it's also called outside function.

Unix/scripts/config_keytab_update.sh
```diff
@@ -69,11 +78,13 @@ do
     case "$1" in
 	--configure)
         configure
+        cleanTempFile
 	    shift 1
 	    ;;
 
 	--unconfigure)
         unconfigure
+        cleanTempFile
```
palladia: Probably better to call cleanTempFile after this do loop, so it will cover the error case also.

Unix/scripts/config_keytab_update.sh
JumpingYang001: good catch, will fix it, thanks.

Unix/scripts/config_keytab_update.sh
JumpingYang001: great! will fix it, thanks.

JumpingYang001:
@palladia , I have fixed the issue and tested the tmp files issue has gone, could you help to approve it? thanks.

PR #572, opened 2018-09-06
Changes to make 64-bit omi pacage for HPUX

sarojcare:
Changes to create HPUX 64-bit omi package instead of 32-bit omi package.


JumpingYang001:
already merged to master: https://github.com/Microsoft/omi/commit/6725abfcc71eaeb25ded69c754483d30001fe1b5, so close it, thanks.

PR #581, opened 2018-09-28
Allow writing SPNEGO responses over HTTPS

Iristyle:
- Previously a check was performed when writing headers to see if the current
   handler is operating over SSL. In cases where the message was sent as
   Content-Type: multipart/encrypted, the message response would be returned
   as application/soap+xml if the transport was HTTPS. This could confuse
   clients (such as the winrm gem) as they are expecting a response of
   multipart/encrypted.

   By all accounts, the PSRP implementation on Windows in PowerShell 5 and
   PowerShell 6 allows multipart/encrypted spnego messages to be sent over
   HTTPS.

   Since the code already checks for handler->encryptedTransaction prior to
   determining how to write response headers, it is safe to remove the SSL
   check to allow mulitpart/encrypted over HTTPS. This change should
   simply make the protocol handling more tolerant.

 - This fixes an issue when trying to establish a connection to an OMI server
   from the WinRM Ruby library where NTLM/SPNEGO is enabled in the OMI server.

 - Similarly change the behavior in httpclient.c to match what has been changed
   on the server response writing in this commit, and the server request
   reading in the prior commit ebe43ea5aa48300f222946cb8eaefa7478b204f4

### DIFF
--- a/Unix/http/http.c
+++ b/Unix/http/http.c
@@ -939,68 +939,65 @@ static Http_CallbackResult _WriteHeader( Http_SR_SocketData* handler)
             content_len = handler->sendPage->u.s.size;
         }
 
-        if (!handler->ssl)
-        {
     #if ENCRYPT_DECRYPT
-            if (handler->encryptedTransaction)
+        if (handler->encryptedTransaction)
+        {
+            Page *pOldPage = handler->sendPage;
+
+            if (!Http_EncryptData(handler, content_len, content_type_len, content_type, &handler->sendPage) )
             {
-                Page *pOldPage = handler->sendPage;
-        
-                if (!Http_EncryptData(handler, content_len, content_type_len, content_type, &handler->sendPage) )
+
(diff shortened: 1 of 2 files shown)


Iristyle:
Like PR #577, this removes `->ssl` guards when sending responses to clients. Clients that are sending `Content-Type: multipart/encrypted` messages expect to receive responses in the same manner.

I was able to get a fully fleshed out PR (with tests that include real messages from OMI server) up against the winrm gem to fix the interop issues I encountered at https://github.com/WinRb/WinRM/pull/288

I've traced the message flow in WireShark and with this patch, I see the expected responses being generated by the server - and the client is able to decrypt / process the messages correctly. I've additionally tested using `Enter-PSSession` from a Windows host with PowerShell 6.1.0 and found everything to behave correctly.  This is probably something that should have some tests, but I figured I would leave that up to the MS team.


Thanks for having a look @paulcallen @yakman2020 @JumpingYang001

PR #586, opened 2018-10-11
update READ.me

BingyuLiu:


JumpingYang001:
Hi John, the commit message suggest to use like 'update support matrix in readme'.

our final OMI/SCX UR support matrix is:
1. remove CentOS 5 (x86 and x64)
2. remove debian 6, 7 (x86 and x64)
3. remove suse10 (x86 and x64)
4. add suse 12 ppc
5. add suse 15

Unix:
1. remove aix 6.1
2. aix 7.1: binary from 32 bit to 64 bit.
3. HPUX: binary from 32 bit to 64 bit.

BingyuLiu:
thanks Jumping.

PR #594, opened 2018-10-22
Improves omid.service when systemd not ready

JumpingYang001:
@microsoft/omi-devs, found a systemd configuration issue when systemd process crash.

When systemd crash:
1.Expect omi installation logic: Systemd logic should copy the omid.systemd to omid.service and reload it and enable it.
2.Current omi installation logic: In fact, my previous fix and case go to use line 191-192: 
https://github.com/Microsoft/omi/blob/master/Unix/installbuilder/datafiles/Linux.data#L191-L192

My previous works fine for the scenario, because systemd load /etc/init.d/omid, but it should load /lib/systemd/system/omid.service even the wrong logic works.

### DIFF
--- a/Unix/installbuilder/datafiles/Linux.data
+++ b/Unix/installbuilder/datafiles/Linux.data
@@ -50,14 +50,7 @@ ResolveSystemdPaths()
 {
     local UNIT_DIR_LIST="/usr/lib/systemd/system /lib/systemd/system"
 
-    pidof systemd 1> /dev/null 2> /dev/null
-    if [ $? -eq 0 ]; then
-        # Be sure systemctl lives where we expect it to
-        if [ ! -f /bin/systemctl ]; then
-            echo "FATAL: Unable to locate systemctl program" 1>&2
-            exit 1
-        fi
-
+    if [ -d /run/systemd/system ]; then
         # Find systemd unit directory
         for i in ${UNIT_DIR_LIST}; do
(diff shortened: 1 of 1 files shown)


JumpingYang001:
found ResolveSystemdPaths will check systemd process, have to update the code again, updating, in fact, I would like to remove all systemd process active checking, because we don't care it, we just make sure `/bin/systemctl` exist and omi will work with systemd system even sysyemd process crash.

JumpingYang001:
Updated, but not remove "systemd process checking" logic, I just keep them now, if it is good or necessary to remove the "systemd process checking" logic, I can remove them, thanks.
I will test related cases tomorrow, thanks.

JumpingYang001:
@palladia @sarojcare can you help to approve this fix? thank you!

JumpingYang001:
@sarojcare can you help to approve this fix? thank you!

sarojcare:
Do we need the lines hilighted. Is it ok to use systemctl as directly?       
 **if [ -x /bin/systemctl ]; then
               /bin/systemctl stop ${SERVICE}
            else**
               systemctl stop ${SERVICE}
            **fi**

sarojcare:
Do you mean on those platforms if you just run systemctl it will not able to find it's location?

- Saroj

JumpingYang001:
yes

JumpingYang001:
ok, will share you.

PR #601, opened 2018-11-15
malloc SDL items assigned to George

palladia:


BingyuLiu:
I added comments several weeks ago, but I forgot to click the "Submit review" button. So George can't see my comments.

This code change looks good to me.

PR #604, opened 2018-11-28
With limitation: SDL items assigned to John

BingyuLiu:


JumpingYang001:
duplicate with #603 ?

BingyuLiu:
> duplicate with #603 ?

It is different, #603 has not a size limitation.

BingyuLiu:
@JumpingYang001 , In some OS, the net.core.rmem_max  is 212992 by default. User can change it by modifying /etc/sysctl.conf file. I am not sure 50M is suitable, that is why I made 2 code changes 603 and 604.

palladia:
Your new code doesn't even compile cleanly.  Please compile and test before submitting for review.

Issue #606, opened 2018-12-11
Add support for FreeBSD


wfurt:
We would like to improve monitoring on Azure VMs running FreeBSD. It seems like the Azure agent sits on top of OMI.

PR #607, opened 2018-12-11
initial freebsd support

wfurt:
related to #606. This is initial support without packaging or documentation. 
With this change it is possible build and run omiserver and use cli.

I'm able to run all tests.
```
./regress
 
##===================##
##                                            ##
## ALL TESTS PASSED!        ##
##                                            ##
##===================##
 
```

and this on my Azure machine
```
root@toweinfu-fbsd:/opt/omi-1.6.0 # sudo /opt/omi/bin/omicli ei root/omi OMI_Identify
instance of OMI_Identify
{
    [Key] InstanceID=2FDB5542-5896-45D5-9BE9-DC04430AAABE
    SystemName=toweinfu-fbsd
    ProductName=OMI
    ProductVendor=Microsoft
    ProductVersionMajor=1
    ProductVersionMinor=6
    ProductVersionRevision=0
    ProductVersionString=1.6.0-121
    Platform=FREEBSD_X86_64_GNU
    OperatingSystem=FREEBSD
    Architecture=X86_64
    Compiler=GNU
    ConfigPrefix=GNU
    ConfigLibDir=/opt/omi-1.6.0/lib
    ConfigBinDir=/opt/omi-1.6.0/bin
    ConfigIncludeDir=/opt/omi-1.6.0/include
    ConfigDataDir=/opt/omi-1.6.0/share
    ConfigLocalStateDir=/opt/omi-1.6.0/var
    ConfigSysConfDir=/opt/omi-1.6.0/etc
    ConfigProviderDir=/opt/omi-1.6.0/etc
    ConfigLogFile=/opt/omi-1.6.0/var/log/omiserver.log
    ConfigPIDFile=/opt/omi-1.6.0/var/run/omiserver.pid
    ConfigRegisterDir=/opt/omi-1.6.0/etc/omiregister
    ConfigSchemaDir=/opt/omi-1.6.0/share/omischema
    ConfigNameSpaces={interop, root-cimv2, root-omi, root-check}
}
```

### DIFF
--- a/Unix/pal/cpu.c
+++ b/Unix/pal/cpu.c
@@ -21,7 +21,11 @@ __attribute__((constructor)) static void InitializeTimeOffset(void)
     struct tm local;
 
     time_t now = time(NULL);
+    // Calculate GMT and local time
+    gmtime_r(&now, &gmt);
+    localtime_r(&now, &local);
 
+    // calculate difference between local time and GMT
     // Since it is only used on one place, pre-calculate to minutes.
     timezone_calculated = (mktime(&gmt) - mktime(&local)) / -60;
 }


wfurt:
no, I won't @palladia. I don't have permissions. Who would be good person to talk to make this more polished? e.g. fix packaging and documentation. I tried to look around but I want to make sure it would meet expectation.

Unix/pal/cpu.c
```diff
@@ -10,6 +10,23 @@
...
+long timezone_calculated = 0;
+
+__attribute__((constructor)) static void InitializeTimeOffset(void)
+{
+    // Unlike SYSV, BSD does not have timezone with time offset. (It returns char* with name)
+    // We need to calculate it once.
+    struct tm gmt;
+    struct tm local;
+
+    time_t now = time(NULL);
```
JumpingYang001: where is now use?

JumpingYang001:
@wfurt Saroj seems be in vacation today and tomorrow, he might be back next week, let's wait his comments, thanks.

Unix/pal/cpu.c
wfurt: look at CPU_GetLocalTimestampz() @JumpingYang001. On Linux, timezone is  extern long with offset to GMT. BSD does not have that. (or OSX/MacOS) Since I failed to figure out what we do for Mac, I simply calculate once difference between GMT and local.

JumpingYang001: yeah, I found there is a [char* timezone() function](https://github.com/freebsd/freebsd/blob/master/include/time.h#L197) in FreeBSD.

I think OSX should use exist logic, I don't know much about how BSD dealt with it.
I just surprise the `now` variable here isn't used?
:)

JumpingYang001: I am not sure if my condition is good, just a  thought. :)

Issue #612, opened 2019-01-14
container and win10 Ubuntu18 bash installation issue since omi 1.4.3-1


JumpingYang001:
we found it in recently testing: I will new a PR to fix this issue.
```
root@ost64-ub18-03:~# dpkg -i omi-1.8.0-102.ssl_110.ulinux.x64.deb
Selecting previously unselected package omi.
(Reading database ... 144771 files and directories currently installed.)
Preparing to unpack omi-1.8.0-102.ssl_110.ulinux.x64.deb ...
/var/lib/dpkg/tmp.ci/preinst: 134: /var/lib/dpkg/tmp.ci/preinst: /opt/omi/bin/service_control: not found
Creating omiusers group ...
Creating omi group ...
Creating omi service account ...
Unpacking omi (1.8.0.102) ...
Setting up omi (1.8.0.102) ...
Generating a 2048 bit RSA private key
....................................+++
....................................+++
writing new private key to '/etc/opt/omi/ssl/omikey.pem'
-----
2019-01-14 05:08:56 : Crontab no longer configured to update omi keytab.
2019-01-14 05:08:56 System already configured to run /opt/omi/bin/support/config_keytab_update.sh automatically
Checking if cron is installed...
Checking if cron/crond service is started...
Set up a cron job to OMI logrotate every 15 minutes
Processing triggers for systemd (237-3ubuntu10.9) ...

Processing triggers for ureadahead (0.100.0-20) ...
```

JumpingYang001:
Fixed.

PR #613, opened 2019-01-14
Fix service_control not found in container

JumpingYang001:
Fixes issue #612, @Microsoft/omi-devs 
The issue has added in #495


JumpingYang001:
@sarojcare can you help to approve this fix? thank you!

PR #599, opened 2018-11-07
Fix login.cfg in AIX

JumpingYang001:
@Microsoft/omi-devs , fixes #598 , could you help to review this PR? thanks very much!

PR #616, opened 2019-01-18
ensure /bin/false is added in 'shells' value in login.cfg

BingyuLiu:
Fix "Error changing "shell" to "/bin/false" : Value is invalid."


JumpingYang001:
Duplicated with #599 ?

JumpingYang001:
#599 will also remove exist problem on customer's env, this PR doesn't consider this scenario, thanks.

BingyuLiu:
yes, duplicated. I will give up my code change. :)

BingyuLiu:
close it, since Jumping has fixed.

PR #639, opened 2019-07-19
Change "-e" to "-f" to fix Solaris install failure

JumpingYang001:
On Solaris, omi install fail as below message, after change "-e" to "-f", the issue has gone.
```
## Installing part 1 of 1.
[ verifying class <none> ]
[ verifying class <config> ]
## Executing postinstall script.
/var/sadm/pkg/MSFTomi/install/postinstall: test: argument expected
pkgadd: ERROR: postinstall script did not complete successfully
Installation of <MSFTomi> failed.

```

### DIFF
--- a/Unix/installbuilder/datafiles/Base_OMI.data
+++ b/Unix/installbuilder/datafiles/Base_OMI.data
@@ -370,7 +370,7 @@ OPENSSL_PATH="openssl"
 #endif
 
 is_suse11_platform_with_openssl1(){
-  if [ -e /etc/SuSE-release ];then
+  if [ -f /etc/SuSE-release ];then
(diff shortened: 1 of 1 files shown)


sarojcare:
Move the changes related to suse11 in commit c4b919f from Base_OMI.data to Linux_RPM.data instead.

JumpingYang001:
@sarojcare we failed at generate certs before the suse11 fixes, if we want to move to Linux_RPM.data, we also need to move generate cert logic to other Unix parts data files, how do you think?
`GenerateKeyCert` call `$OPENSSL_PATH` :https://github.com/microsoft/omi/blob/master/Unix/installbuilder/datafiles/Base_OMI.data#L403

PR #649, opened 2019-10-24
Add IPv6 support

JumpingYang001:
@microsoft/omi-devs  fixes ipv6 doesn't listen on omiserver.

### DIFF
--- a/Unix/http/http.c
+++ b/Unix/http/http.c
@@ -1702,11 +1702,20 @@ static MI_Result _CreateAddListenerSocket(
 
     /* Create listener socket */
     {
-        Addr_InitAny(&addr, port);
+        /* Try IPv6 at first */
+        trace_Trying_IPv6();
+        Addr_InitAnyIPv6(&addr, port);
         r = Sock_CreateListener(&listener, &addr);
-
         if (r != MI_RESULT_OK)
         {
+            /* When IPv6 fail, try IPv4. */
+            trace_Trying_IPv4();
+            Addr_InitAny(&addr, port);
+            r = Sock_CreateListener(&listener, &addr);
+        }
+        if (r != MI_RESULT_OK)
+        {
+            trace_Listen_Failed();
             return r;
         }
 
--- a/Unix/base/oi_traces.h
+++ b/Unix/base/oi_traces.h
@@ -907,6 +907,16 @@ OI_EVENT("OMI reloadConfig completed")
 void trace_ReloadConfig_OMI_OK();
 OI_EVENT("OMI reloadDispatcher completed")
 void trace_ReloadDispatcher_OMI_OK();
+OI_EVENT("Trying to listen on IPv4")
+void trace_Trying_IPv4();
+OI_EVENT("Trying to listen on IPv6")
+void trace_Trying_IPv6();
+OI_EVENT("Listen failed on both IPv4 and IPv6")
+void trace_Listen_Failed();
+OI_EVENT("Turn off IPV6_V6ONLY failed.")
+void trace_TurnOff_IPV6_V6ONLY_Failed();
+OI_EVENT("Turn off IPV6_V6ONLY pass.")
+void trace_TurnOff_IPV6_V6ONLY_Pass();
 
 
 /******************************** INFORMATIONAL ***********************************/
--- a/Unix/base/oiomi.h
+++ b/Unix/base/oiomi.h
@@ -2453,6 +2453,36 @@ FILE_EVENT0(30219, trace_ReloadConfig_OMI_OK_Impl, LOG_WARNING, PAL_T("OMI reloa
 #endif
 FILE_EVENT0(30220, trace_ReloadDispatcher_OMI_OK_Impl, LOG_WARNING, PAL_T("OMI reloadDispatcher completed"))
 #if defined(CONFIG_ENABLE_DEBUG)
+#define trace_Trying_IPv4() trace_Trying_IPv4_Impl(__FILE__, __LINE__)
+#else
+#define trace_Trying_IPv4() trace_Trying_IPv4_Impl(0, 0)
+#endif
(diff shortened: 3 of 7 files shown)


Unix/http/http.c
```diff
@@ -1702,11 +1702,20 @@ static MI_Result _CreateAddListenerSocket(
 
     /* Create listener socket */
     {
-        Addr_InitAny(&addr, port);
+        /* Try IPv6 at first */
+        Addr_InitAnyIPv6(&addr, port);
```
bbsaurav: shouldn't we be listening on both IPv4 and IPv6?
Devices can have both IPv4 and IPv6 connectivity giving preference to IPv6 will impact IPv4 connections.

JumpingYang001: listen on IPv6  will also accept ipv4 and ipv6 connections, check this: https://www.ibm.com/support/knowledgecenter/ssw_ibm_i_72/rzab6/xacceptboth.htm

JumpingYang001:
@sarojcare how about `EnableIPv6=true` in omiserver.conf?
will do as below.
```
By Default look for ipv4 if fails try with ipV6.
In case in config it is mentioned to use IPv6 then try binding with IPv6 only else fail gracefully with appropriate message.
```

Unix/http/http.c
JumpingYang001: I mean `in6addr_any` will listen on both IPv6 and IPv4.
```
/********************************************************************/
      /* After the socket descriptor is created, a bind() function gets a */
      /* unique name for the socket.  In this example, the user sets the  */
      /* address to in6addr_any, which (by default) allows connections to */
      /* be established from any IPv4 or IPv6 client that specifies port  */
      /* 3005. (that is, the bind is done to both the IPv4 and IPv6 TCP/IP    */
      /* stacks).  This behavior can be modified using the IPPROTO_IPV6   */
      /* level socket option IPV6_V6ONLY if required.                      */
      /********************************************************************/
````

**Comment by JumpingYang001** [comment]
@sarojcare  I tried on only enabled IPv6 box cent6x64testl06, IPv4 listen pass even it doesn't have IPv4 address. so it will not try IPv6.

File: Unix/http/http.c (hunk 1 above)
Comment (JumpingYang001) [reply Unix/http/http.c]: I need to learn and do some investigation for `in6addr_any` http://man7.org/linux/man-pages/man7/ipv6.7.html
```
Linux 2.2 optionally implements the Internet Protocol, version 6.
       This man page contains a description of the IPv6 basic API as
       implemented by the Linux kernel and glibc 2.1.  The interface is
       based on the BSD sockets interface; see socket(7).

       The IPv6 API aims to be mostly compatible with the IPv4 API (see
       ip(7)).  Only differences are described in this man page.

       To bind an AF_INET6 socket to any process, the local address should
       be copied from the in6addr_any variable which has in6_addr type.  In
       static initializations, IN6ADDR_ANY_INIT may also be used, which
       expands to a constant expression.  Both of them are in network byte
       order.

       The IPv6 loopback address (::1) is available in the global
       in6addr_loopback variable.  For initializations,
       IN6ADDR_LOOPBACK_INIT should be used.

       IPv4 connections can be handled with the v6 API by using the
       v4-mapped-on-v6 address type; thus a program needs to support only
       this API type to support both protocols.  This is handled
       transparently by the address handling functions in the C library.

       IPv4 and IPv6 share the local port space.  When you get an IPv4
       connection or packet to an IPv6 socket, its source address will be
       mapped to v6 and it will be mapped to v6.
```

Comment (bbsaurav) [reply Unix/http/http.c]: can you check the behavior in some older OS like Fedora 24, Ubuntu 14 or 16 and see if listening to IPv6 only accepts connections for both IPv4 and IPv6?

Comment (JumpingYang001) [reply Unix/http/http.c]: will check old OSs.

**Comment by JumpingYang001** [comment]
@sarojcare @bbsaurav how about below design? it will keep current IPv4 by default, if someone want to use IPv6 he need to set omiserver.conf.
BTW, if IPv4 socket listen on 1270, IPv6 socket seems cannot listen on 1270 due to conflicts. If they use `in6addr_any` will not have the problem for scenarios "IPv6 only" and "IPv6+IPv4".

By default, enabledIPv6=false in omiserver.conf.
```
if(enableIPv6)
{
    TryIPv6();
    if(tryFail) TryIPv4();
}
else
{
    TryIPv4();
    if(tryFail) TryIPv6();
}
```

bbsaurav:
> @sarojcare @bbsaurav how about below design? it will keep current IPv4 by default, if someone want to use IPv6 he need to set omiserver.conf.
> BTW, if IPv4 socket listen on 1270, IPv6 socket seems cannot listen on 1270 due to conflicts. If they use in6addr_any will not have the problem for scenarios "IPv6 only" and "IPv6+IPv4".
Can you check how sshd listens on port 22 for both IPv4 and IPv6
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN      63615/sshd
tcp6       0      0 :::22                   :::*                    LISTEN      63615/sshd

For some functionality to work we use ssh communication between MS and monitored device.

> By default, enabledIPv6=false in omiserver.conf.
> if(enableIPv6)
> {
>     TryIPv6();
>     if(tryFail) TryIPv4();
> }
> else
> {
>     TryIPv4();
>     if(tryFail) TryIPv6();
> }
My only concern here is that when we only listen on IPv6 then for IPv4 communication the IPv4 address will be translated into IPv6 address. In most of the recent OS this translation layer is already present. But in older OS this translation layer might not be present. You need to check whether this translation layer is present on all the supported Linux and Unix distros and their versions.

JumpingYang001:
@bbsaurav got you, some things need to do:
1. setup only IPv6 environments on all supported Linux and Unix.
2. kick off a private on Jenkins.
3. test if works.

JumpingYang001:
Deepak merged #655 , so close this PR.

PR #651, opened 2019-11-08
Removed OS/X build location searching script

snchennapragada:


Unix/installbuilder/buildNuget.sh
```diff
@@ -105,12 +105,12 @@ EOF
 # Note, up above, that we are now in the installbuilder/nuget directory
 #
 
-mkdir -p runtimes/linux-x64/native runtimes/osx/native
```
JumpingYang001: `mkdir -p runtimes/linux-x64/native` need to keep. and only remove `runtimes/osx/native `
otherwise `cp ${build_linux}/libmi.so runtimes/linux-x64/native` will fail.

snchennapragada:
Script is approved.

PR #657, opened 2020-03-28
Don't delete and re-create the user 'omi' if it is already exist

deepakjain111:


sarojcare:
Is there any dependency on omi group? Do we need to make validation that omi user existing is also part of omi group?

deepakjain111:
> Is there any dependency on omi group? Do we need to make validation that omi user existing is also part of omi group?

I don't think so. We don't need that validation.

sarojcare:
Whether pbuild and unit testing has done? 

Also the comment line should not be longer than 70 chars. In case we need to have longer string please put it in a file with each line not having 70 chars and use it to commit.

deepakjain111:
> Whether pbuild and unit testing has done?
> 
> Also the comment line should not be longer than 70 chars. In case we need to have longer string please put it in a file with each line not having 70 chars and use it to commit.

I have done unit testing and will ask testing team to test on all Linux system

PR #670, opened 2020-06-02
Fix up GSSAPI auth on macOS

jborean93:
The current code has a few problems when running on macOS

* The `HEIMDAL` macro was not set on macOS, this causes it to try and load the wrong names.
* `GSS_C_NT_HOSTBASED_SERVICE` is not used anywhere in the code (although it probably should be)
    * Planning on revisiting this in another PR as this is the recommended name type for the target name
    * In any case the name is invalid on macOS' Heimdal implementation so just removing it was easier for now
* `GSS_KRB5_NT_PRINCIPAL_NAME` is valid for MIT krb5 but on Heimdal it needs to be accessed with `__gss_krb5_nt_principal_name_oid_desc`

Finally there was an issue when using an IP that was not routable. The `getaddrinfo()` function would set a `NULL` value for `ai_canonname`. This causes a seg fault when calling `strlen(info->ai_canonname)`. Instead this just falls back to using the HTTP hostname that was set and let it fail later on.

Ultimately this enables Kerberos auth to work on macOS.

Unfortunately enabling NTLM auth through SPNEGO requires a lot more effort that I still need to investigate. Right now the code sends the NTLM negotiate and processes the challenge message but it fails to send the authenticate message causing a failure. Hopefully it's a simple fix I can address in another PR.

### DIFF
--- a/Unix/buildtool
+++ b/Unix/buildtool
@@ -923,6 +923,7 @@ if [ "$arg1" = "cflags" -o "$arg1" = "cxxflags" ]; then
             r="$r -Wall"
             r="$r -I/usr/local/include"
             r="$r -Dmacos"
+            r="$r -DHEIMDAL"
             r="$r -fvisibility=hidden"
             r="$r -fstack-protector-all"
             r="$r -DGSS_USE_IOV=1"
--- a/Unix/http/httpclientauth.c
+++ b/Unix/http/httpclientauth.c
@@ -290,7 +290,6 @@ typedef struct _Gss_Extensions
 #endif
 
     gss_OID Gss_Nt_Service_Name;
-    gss_OID Gss_Nt_HostBased_Service_Name;
     gss_OID Gss_Krb5_Nt_Principal_Name;
     gss_OID Gss_C_Nt_User_Name;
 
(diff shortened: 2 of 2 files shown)


JumpingYang001:
thanks for PR! this PR seems be not special to MacOS, and it seems a new feature about add `HEIMDAL` support while we currently used MIT Kerberos. And any new feature needs to be trigger by OMI PM/Manager internal and there is some process internal, if you want to do this new feature, the best way is to have a customer request a DCR internal, and then there might be some process to trigger it, PM will determine trigger it or not internal based on business justification, etc. just like that.

jborean93:
> this PR seems be not special to MacOS

I somewhat disagree, macOS uses Heimdal so fixing one fixes the other.

> And any new feature needs to be trigger by OMI PM/Manager internal and there is some process internal

Disappointing that a "open source" project still needs to have an internal review that is completely opaque to contributors. You may as well just remove the repo and keep everything internal.

Ultimately this is maintained by some team in Microsoft. If they don't want fixes for their code that's contributed by outsiders then good luck to them. Feel free to take whatever you want from here in the future if you desire. I'll just maintain my own fork for my own purposes.

PR #692, opened 2021-08-09
Enhanced security

deepakjain111:
Introduced authentication for omiserver in omiengine for user validation and PAM based user validation.
This authentication is secret string based. This secret string can be set only once, that is duirng omiegine bootup, any other afterwards  attempt to change the secret string will be ignored.

Default user type for running any scripts will be invalid user, so that user authencication can't be skipped. Additional validation for not allowing unauthorized user to proceed.

Fixed a Use-After-Free vulnerability where omiengine's connection with client get freed before omiegnine's connection with omiserver amd resulting in Use-After-Free vulnerability.


JumpingYang001 (approved):
LGTM.

PR #735, opened 2023-02-13
Setting hdl inside AUTHORIZATION macro as it is defined inside the macro

ganeshkamath89:
Setting hdl inside AUTHORIZATION macro as it is defined inside the macro

Note: If user disables AUTHORIZATION macro this code gives compilation error


ganeshkamath89:
@JumpingYang001, seems to me like there is some issue with Travis CI. Can you please review and check once?

